PR #1, opened 2025-02-07
Payments contract first draft

aarshkshah1992:
This is the first draft for the payments contract MVP. It has the following features: 

- ERC-20 deposits
- ERC-20 withdrawals
- rail creation
- operator approvals
- operator termination
- rail payment modification
- rail lockup modification
- lazy rail settlement
- proxy pattern for contract upgrade
- arbitration
- rate change queuing
- avoiding debt creation during lockup or rate modification

All rail modifications and creation respect the operator approval.

The following features which will be added in a subsequent PR:

1) Escrow
2) Payment Fees


README.md
```diff
@@ -1,3 +1,66 @@
-# payments
+## Foundry
```
Stebalien, line 1: Revert this change?

src/Payments.sol
```diff
@@ -0,0 +1,418 @@
...
+            // Update operator's used base
+            approval.base_used += once;
+        }
+
+        // Update payer's lockup rate and base
+        payer.lockupBase = payer.lockupBase - (oldRate * rail.lockupPeriod) + (rate * rail.lockupPeriod);
+        payer.lockupRate = payer.lockupRate - oldRate + rate;
+
+        // Init hasLockupStarted flag and lockupStart for the first rail with non-zero rate
+        if (rate > 0 && !payer.hasLockupStarted) {
```
Stebalien, line 341: Take a look at the "Lockup can be settled..." part of: https://www.notion.so/filecoindev/Payments-Design-Iteration-17ddc41950c180468806c7e0097c2e14?pvs=4#185dc41950c1805da8d4ce567e343307

`lockupStart` is probably misnamed. Really, it means "the last time we settled lockup". I.e., it's when the lockup rate starts applying.

My point is, I don't think we need `hasLockupStarted`. We always adjust `lockupStart`.

Stebalien: Ah, I see you've implemented `accumulateLockupRate`. You don't need this code here, you just need to call `accumulateLockupRate` before changing the rate (e.g., at the very top of this function). The first time this is called, that will accumulate nothing because the rate will be 0, but it will initialize `lockupStart`.

src/Payments.sol
```diff
@@ -0,0 +1,572 @@
...
+        require(from != address(0), "from address cannot be zero");
+        require(to != address(0), "to address cannot be zero");
+        require(operator != address(0), "operator address cannot be zero");
+
+        OperatorApproval memory approval = operatorApprovals[token][from][operator];
+        require(approval.isApproved, "operator not approved");
+
+        Account storage toAccount = accounts[token][to];
+        require(toAccount.ownerAddress != address(0), "to account does not exist");
+        require(toAccount.funds > 0, "to account has no funds");
```
Stebalien, line 220: They don't need funds, do they?

src/Payments.sol
```diff
@@ -0,0 +1,572 @@
...
+
+        Account storage fromAccount = accounts[token][from];
+        require(fromAccount.ownerAddress != address(0), "from account does not exist");
+        require(fromAccount.funds > 0, "from account has no funds");
```
Stebalien, line 224: Technically, the sender doesn't need any funds either.

src/Payments.sol
```diff
@@ -0,0 +1,572 @@
...
+
+        if (approval.arbitrer != address(0)) {
+            require(arbiter == approval.arbitrer, "arbiter mismatch");
+        }
```
Stebalien, line 228: TBD: need to discuss approved operators and who, precisely, is allowed to call `createRail`.

src/Payments.sol
```diff
@@ -0,0 +1,572 @@
...
+
+        uint256 railId = _nextRailId++;
+
+        Rail storage rail = rails[railId];
```
Stebalien, line 232: OT: Ethereum's "everything exists, it just defaults to zero" gets me every time. It's kind of cool but also really disconcerting.

src/Payments.sol
aarshkshah1992: @Stebalien I mean not strictly speaking. We can get rid of this yeah.

src/Payments.sol
aarshkshah1992: @Stebalien Technically no but I'd like to keep this so that rails are only created with clients who have interacted with the contract before. Just think it's better that way.

src/Payments.sol
aarshkshah1992: @Stebalien What's "OT" ?

.gitignore
```diff
@@ -0,0 +1,14 @@
+# Compiler files
+cache/
+out/
+
+# Ignores development broadcast logs
+!/broadcast
+/broadcast/*/31337/
+/broadcast/**/dry-run/
```
Stebalien, line 8: Why not just ignore all of /broadcast?

src/Payments.sol
```diff
@@ -0,0 +1,572 @@
...
+import "@openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol";
+import "@openzeppelin/contracts-upgradeable/proxy/utils/UUPSUpgradeable.sol";
+import "@openzeppelin/contracts-upgradeable/access/OwnableUpgradeable.sol";
+import "./RateChangeQueue.sol";
+
+interface IArbiter {
+    struct ArbitrationResult {
+        bool approved;
+        uint256 modifiedAmount;
+        uint256 settledUntil;
```
Stebalien, line 16: We should consider adding an arbitrary "note" field. Could be useful to explain why the full amount isn't approved.

src/Payments.sol
```diff
@@ -0,0 +1,572 @@
...
+import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+import "@openzeppelin/contracts/utils/ReentrancyGuard.sol";
+import "@openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol";
+import "@openzeppelin/contracts-upgradeable/proxy/utils/UUPSUpgradeable.sol";
+import "@openzeppelin/contracts-upgradeable/access/OwnableUpgradeable.sol";
+import "./RateChangeQueue.sol";
+
+interface IArbiter {
+    struct ArbitrationResult {
+        bool approved;
```
Stebalien, line 14: Do we need `approved`? IMO, we should just revert if we're not approving anything, but I'm not really a solidity expert.

src/Payments.sol
```diff
@@ -0,0 +1,572 @@
...
+
+contract Payments is Initializable, UUPSUpgradeable, OwnableUpgradeable, ReentrancyGuard {
+    using SafeERC20 for IERC20;
+    using RateChangeQueue for RateChangeQueue.Queue;
+
+    struct Account {
+        address ownerAddress;
+        uint256 funds;
+        uint256 lockupCurrent;
+        uint256 lockupRate;
```
Stebalien, line 35: Let's comment on the units here (and all the other fields). E.g. lockup rate is attotoken/epoch.

src/Payments.sol
```diff
@@ -0,0 +1,572 @@
...
+        bool isActive;
+        address token;
+        address from;
+        address to;
+        address operator;
+        address arbiter;
+        uint256 paymentRate;
+        uint256 lockupPeriod;
+        uint256 lockupFixed;
+        uint256 lastSettledAt;
```
Stebalien, line 49: Hm. Now that I see this, I wonder if we should call this "settledUpTo" instead of "lastSettledAt".

src/Payments.sol
```diff
@@ -0,0 +1,572 @@
...
+    }
+
+    struct OperatorApproval {
+        bool isApproved;
```
Stebalien, line 53: Do we need this field?

src/Payments.sol
```diff
@@ -0,0 +1,572 @@
...
+        address arbitrer;
```
Stebalien, line 54: Do we need this field either? Isn't it implicit?

src/Payments.sol
Stebalien: Well, this goes back to who can create a rail. IMO, the answer is: the client.

It would be nice if a client can go through the entire signup flow without having to deposit any funds until they actually start using the service.

src/Payments.sol
```diff
@@ -0,0 +1,572 @@
...
+
+
+        finalEpoch = currentEpoch;
+        return (totalSettled, finalEpoch);
+    }
+
+    /// @notice Returns the minimum of two numbers.
+    function min(uint256 a, uint256 b) internal pure returns (uint256) {
+        return a < b ? a : b;
+    }
```
Stebalien, line 530: Make this a free function?

src/Payments.sol
```diff
@@ -0,0 +1,572 @@
...
+
+            // Short circuit and return if _settleSegment returns an epoch less than what we sent it
+            if (segmentSettledEpoch < nextBoundary) {
+                return (totalSettled + segmentAmount, segmentSettledEpoch);
+            }
+
+            totalSettled += segmentAmount;
+            currentEpoch = segmentSettledEpoch;
+            rateQueue.dequeue();
+            }
```
Stebalien, line 520:
```suggestion
        }
```

src/Payments.sol
```diff
@@ -0,0 +1,572 @@
...
+            settledUntil = arbResult.settledUntil;
+            settledAmount = arbResult.modifiedAmount;
+        }
+
+        // If no progress is made, revert.
+        require(settledUntil > segmentStart, "no progress made in settlement");
+
+        return (settledAmount, settledUntil);
+    }
+    function _settleWithRateChanges(
```
Stebalien, line 477:
```suggestion
    }

    function _settleWithRateChanges(
```

src/Payments.sol
```diff
@@ -0,0 +1,572 @@
...
+
+        // settle account lockup and if account is not funded upto to the current epoch; revert
+        (bool funded, uint256 settleEpoch) = settleAccountLockup(payer);
+        require(funded && settleEpoch == block.number, "cannot modify lockup as client does not have enough funds for current settlement");
+
+        // Calculate the change in base lockup
+        uint256 oldLockup = rail.lockupFixed + (rail.paymentRate * rail.lockupPeriod);
+        uint256 newLockup = lockupFixed + (rail.paymentRate * period);
+
+        // assert that operator allownace is respected
```
Stebalien, line 265:
```suggestion
        // assert that operator allowance is respected
```

src/Payments.sol
```diff
@@ -0,0 +1,572 @@
...
+        if (rail.arbiter != address(0)) {
+            // Call the external arbitrator.
+            IArbiter arbiter = IArbiter(rail.arbiter);
+            IArbiter.ArbitrationResult memory arbResult = arbiter.arbitratePayment(
+                railId,
+                settledAmount,
+                segmentStart,
+                segmentEnd
+            );
+            require(arbResult.approved, "arbitrer refused payment");
```
Stebalien, line 466: We need to validate that:

1. `arbResult.settledUntil <= settledUntil`.
2. `arbResult.modifiedAmount <= rate * (arbResult.settledUntil - segmentStart)`.

src/Payments.sol
```diff
@@ -0,0 +1,572 @@
...
+        )
+    {
+        // Get the rail and the involved accounts.
+        Rail storage rail = rails[railId];
+        Account storage payer = accounts[rail.token][rail.from];
+        Account storage payee = accounts[rail.token][rail.to];
+
+        settleAccountLockup(payer);
+
+        uint256 settlementTargetEpoch = min(untilEpoch, payer.lockupLastSettledAt + rail.lockupPeriod);
```
Stebalien, line 400: Need to check that `untilEpoch` is at or before the current epoch.

src/Payments.sol
```diff
@@ -0,0 +1,572 @@
...
+        while (currentEpoch < targetEpoch) {
+            // Determine the next boundary: either the next queued rate–change or the targetEpoch.
+            uint256 nextBoundary = targetEpoch;
+            uint256 segmentRate = currentRate;
+            if (!rateQueue.isEmpty()) {
+                RateChangeQueue.RateChange memory nextChange = rateQueue.peek();
+                if (nextChange.untilEpoch > currentEpoch && nextChange.untilEpoch <= targetEpoch) {
+                    nextBoundary = nextChange.untilEpoch;
+                    segmentRate = nextChange.rate;
+                }
```
Stebalien, line 497: I've spent a lot of time staring at this code and I'm still not sure if it's right. Let's talk through it tomorrow.

src/Payments.sol
aarshkshah1992: @Stebalien This is for the Arbitrer contract to gracefully let us know that it is/is not rejecting the payment. The Payment contract does revert the settlement if it sees a disapproval from the arbitrer.

aarshkshah1992: The other option is to not gracefully handle any reverts from the `Arbitrer` contract at all and simply revert right away (in Solidity, when contract X calls contract Y and contract Y reverts -> the call to X reverts right away unless you handle it as above).

The way we've implemented it here -> any revert from `Arbitrer` will bubble up right away and the settlement will revert but atleast for payment refusals -> we get to handle it gracefully and do more stuff if we want to before we revert.

src/Payments.sol
aarshkshah1992: Added this and also now returning this to the caller of `settleRail`

src/Payments.sol
aarshkshah1992: @Stebalien Yeah it's the easiest way to track the fact that an operators has been disapproved. Solidity HashMap/Array deletes aren't clean/gas cheap to do.

src/Payments.sol
aarshkshah1992: @Stebalien My idea was to have the client have a say in who gets to be an arbitrer.

src/Payments.sol
```diff
@@ -0,0 +1,572 @@
...
+    }
+
+    // railId => RateChangeQueue
+    mapping(uint256 => RateChangeQueue.Queue) public railRateChangeQueues;
+
+    // Counter for generating unique rail IDs
+    uint256 private _nextRailId;
+
+    // token => owner => Account
+    mapping(address => mapping(address => Account)) public accounts;
```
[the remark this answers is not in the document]
aarshkshah1992: @Stebalien 

Can't we implement those getters as view functions that can be invoked off-chain on RPC providers ? We can add some state to the contract that makes those view functions slightly less compute intensive but IMO making it a view function that can be invoked off-chain means we can offload the worrying about efficiency to client <> RPC provider relationships without having to add too much state to the contract.

src/Payments.sol
aarshkshah1992: @Stebalien Yeah plan is to raise a subsequent PR for comprehensive docs for both state and functions once we've locked in a meaningful draft.

src/Payments.sol
```diff
@@ -0,0 +1,572 @@
...
+        address owner = accounts[token][msg.sender].ownerAddress;
+        require(owner != address(0), "account does not exist");
+        require(owner == msg.sender, "not account owner");
+        _;
+    }
+
+    function approveOperator(
+        address token,
+        address operator,
+        address arbiter,
```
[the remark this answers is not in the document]
aarshkshah1992: @Stebalien You mean the client creates the rail OR that a client has to approve a rail that has been created by an operator ? Both have different UX implications.

src/Payments.sol
```diff
@@ -0,0 +1,572 @@
...
+        IERC20(token).safeTransfer(msg.sender, amount);
+    }
+
+    function createRail(
+        address token,
+        address from,
+        address to,
+        address operator,
+        address arbiter
+    ) external returns (uint256) {
```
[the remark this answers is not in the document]
aarshkshah1992: @Stebalien It does make the UX harder for clients and operators. It's more intuitive for an operator to create a rail as they know who the provider of the service is etc (without always leaking that to the client).

If we get clients to create a rail -> we need to add a step where the operator has to approve the rail and the operator needs to wait/watch for rail creation before they can start providing services. It really does complicate the entire flow.

src/Payments.sol
aarshkshah1992: I agree. It is involved.

aarshkshah1992: Let me try to simplify it more today before we sync up.

src/Payments.sol
aarshkshah1992: For 1, it should be `arbResult.settledUntil <=segmentEnd`, right ? Yeah making this change.

src/Payments.sol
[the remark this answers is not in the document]
aarshkshah1992: See my comment at https://github.com/FilOzone/payments/pull/1/files#r1961082248.

src/Payments.sol
aarshkshah1992: Done.

src/Payments.sol
aarshkshah1992: Done.

src/Payments.sol
aarshkshah1992: Also renamed the corresponding `ArbitrationResult` field to `settledUpTo`

src/Payments.sol
aarshkshah1992: The Arbitrer SHOULD be able to revert if it wants to.

src/Payments.sol
```diff
@@ -0,0 +1,612 @@
...
+import "@openzeppelin/contracts-upgradeable/proxy/utils/UUPSUpgradeable.sol";
+import "@openzeppelin/contracts-upgradeable/access/OwnableUpgradeable.sol";
+import "./RateChangeQueue.sol";
+
+interface IArbiter {
+    struct ArbitrationResult {
+        bool approved;
+        uint256 modifiedAmount;
+        uint256 settleUpto;
+        string note;
```
aarshkshah1992, line 17: Can leave a note if successful too. For better UX, it's a catch all field for arbitrer to send us info.

src/Payments.sol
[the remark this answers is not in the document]
aarshkshah1992: An operator can have many different arbitrers for different rails.

src/Payments.sol
```diff
@@ -0,0 +1,612 @@
...
+        );
+
+        approval.lockupAllowance = approval.lockupAllowance + oldTotalLockup - newTotalLockup;
+
+        require(newRate <= oldRate + approval.rateAllowance, "new rate exceeds operator rate allowance");
+        approval.rateAllowance = approval.rateAllowance + oldRate - newRate;
+        
+    }
+
+    function updateRailArbiter(uint256 railId, address newArbiter) external validateRailActive(railId) onlyRailOperator(railId) {
```
aarshkshah1992, line 370: Do we really need this ?

src/Payments.sol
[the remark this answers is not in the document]
aarshkshah1992: So remove this.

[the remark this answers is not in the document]
aarshkshah1992: The answer here is to trust the operator.

src/Payments.sol
```diff
@@ -0,0 +1,612 @@
...
+        require(amount <= available, "insufficient unlocked funds for withdrawal");
+        acct.funds -= amount;
+        IERC20(token).safeTransfer(msg.sender, amount);
+    }
+
+    function createRail(
+        address token,
+        address from,
+        address to,
+        address operator,
```
aarshkshah1992, line 208: we don't need this as the `msg.sender` will ALWAYS be the operator.

src/Payments.sol
aarshkshah1992: We agree that neither `from` or `to` needs funds.

src/Payments.sol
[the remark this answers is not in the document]
Stebalien: Ok, we can solve (1) by:

1. Adding a `mapping(address => address[])` of account owners to tokens.
2. Optionally: changing `token => owner => Account` to `owner => token => Account`. This isn't strictly necessary, but it mirrors the mapping in (1).

As discussed, we can drop (2) and let the user handle indirection through smart contracts.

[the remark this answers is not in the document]
Stebalien: Actually, maybe we want a single `owner => Account` mapping where `Account` contains:

1. The client's operator approvals (map of `operator => Approval`).
2. An array of approved client operators.
3. An array of client rail IDs.
4. etc...

src/Payments.sol
```diff
@@ -0,0 +1,612 @@
...
+
+        OperatorApproval storage approval = operatorApprovals[token][msg.sender][operator];
+        approval.arbitrer = arbiter;
+        approval.rateAllowance = rateAllowance;
+        approval.lockupAllowance = lockupAllowance;
+        approval.isApproved = true;
+    }
+
+    // TODO: Revisit
+    function terminateOperator(address operator) external  {
```
aarshkshah1992, line 134: who calls this ?

src/Payments.sol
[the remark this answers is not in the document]
Stebalien: Resolution: we're still exploring the API, let's build that first then figure out what these data structures should look at.

src/Payments.sol
```diff
@@ -0,0 +1,612 @@
+// SPDX-License-Identifier: UNLICENSED
```
aarshkshah1992, line 1: Write down all the `getter` APIs.

aarshkshah1992: Try to make it as easy as possible to access basic state without an Indexer.

aarshkshah1992: Strike a balance between on-chain state and off-chain compute than an RPC provider needs to do to serve Getter requests.

aarshkshah1992: Storing the keys of a Hasmap in an a set if you need iteration. Do it only when needed.

src/Payments.sol
```diff
@@ -0,0 +1,612 @@
...
+
+            OperatorApproval storage approval = operatorApprovals[rail.token][msg.sender][operator];
+            approval.rateAllowance = 0;
+            approval.lockupAllowance = 0;
+            approval.isApproved = false;
+        }
+    }
+
+    // TODO: implement
+    function terminateRail() public {
```
aarshkshah1992, line 164: An operator can always do this by setting everything to 0.

aarshkshah1992: If a client wants to terminate a rail -> they go the operator and get them to modify the rates and lockups for the rail to 0.

src/Payments.sol
```diff
@@ -0,0 +1,612 @@
...
+    struct Account {
+        address ownerAddress;
+        uint256 funds;
+        uint256 lockupCurrent;
+        uint256 lockupRate;
+        uint256 lockupLastSettledAt;
+    }
+
+    struct Rail {
+        bool isActive;
```
aarshkshah1992, line 41: We need this to track rail termination !

src/Payments.sol
```diff
@@ -0,0 +1,612 @@
...
+    returns (
+        uint256 totalSettledAmount,
+        uint256 finalSettledEpoch,
+        string memory note
+    ) {
+        require(untilEpoch <= block.number, "failed to settle: cannot settle future epochs");
+
+        Rail storage paymentRail = rails[railId];
+        Account storage payer = accounts[paymentRail.token][paymentRail.from];
+        Account storage payee = accounts[paymentRail.token][paymentRail.to];
```
aarshkshah1992, line 391: Does the solidity compiler optimise this so we only lookup the map once ?

src/Payments.sol
```diff
@@ -0,0 +1,612 @@
...
+
+        settleAccountLockup(payer);
+
+        // Determine the maximum allowed epoch we can settle to based on the payer's lockupPeriod.
+        uint256 maxSettlementEpoch = min(untilEpoch, payer.lockupLastSettledAt + paymentRail.lockupPeriod);
```
aarshkshah1992, line 396: can read ` payer.lockupLastSettledAt` from return value.

src/Payments.sol
```diff
@@ -0,0 +1,612 @@
...
+
+        // Begin from the last epoch that was actually settled.
+        uint256 lastSettledEpoch = paymentRail.settledUpTo;
+        uint256 currentPaymentRate = paymentRail.paymentRate;
+
+        // Retrieve any queued rate changes for this rail.
+        RateChangeQueue.Queue storage rateQueue = railRateChangeQueues[railId];
```
aarshkshah1992, line 403: We can put the `railRateChangeQueues ` in the `Rail` itself.

src/Payments.sol
```diff
@@ -0,0 +1,612 @@
...
+        require(settledUntil > epochStart, "failed to settle: no progress made in settlement");
+
+        return (settledAmount, settledUntil, note);
+    }
+
+    function _settleWithRateChanges(
+        uint256 railId,
+        Rail storage rail,
+        RateChangeQueue.Queue storage rateQueue,
+        uint256 initialRate,
```
aarshkshah1992, line 509: this is the current rail rate. fix naming.

src/Payments.sol
```diff
@@ -0,0 +1,612 @@
...
+        uint256 startEpoch,
+        uint256 targetEpoch
+    ) internal returns (uint256 totalSettled, uint256 finalEpoch, string memory note) {
+        totalSettled = 0;
+        uint256 currentEpoch = startEpoch;
+        uint256 activeRate = initialRate;
+        note = "";
+
+        while (currentEpoch <= targetEpoch) {
```
aarshkshah1992, line 518: currentEpoch < targetEpoch is what we need here. if it is equal, we are done and we should exit the loop.

src/Payments.sol
```diff
@@ -0,0 +1,612 @@
...
+            uint256 nextBoundary = targetEpoch;
+
+            // If there's an upcoming rate change, check if it applies within our current range.
+            if (!rateQueue.isEmpty()) {
+                RateChangeQueue.RateChange memory upcomingChange = rateQueue.peek();
+                bool isWithinRange = (
+                    upcomingChange.untilEpoch >= currentEpoch &&
+                    upcomingChange.untilEpoch <= targetEpoch
```
Stebalien, line 526: This second condition won't be true if I'm trying to settle up to some historical epoch in the past.

src/Payments.sol
```diff
@@ -0,0 +1,612 @@
...
+
+            // If the arbitration shortened the segment, stop and return immediately.
+            // but keep the rate change in the queue as we've not fully settled the segment.
+            if (segmentEndEpoch < nextBoundary) {
+                return (totalSettled, segmentEndEpoch, arbNote);
+            }
+
+            // Otherwise, we have fully settled up to nextBoundary.
+            currentEpoch = segmentEndEpoch;
+            note = arbNote;
```
Stebalien, line 553: It would be kind of nice to somehow join these notes together when non-empty, but that isn't critical. We mostly care about the last note anyways.

src/Payments.sol
```diff
@@ -0,0 +1,612 @@
...
+
+            // Dequeue the rate change we just passed (if it was indeed used this round).
+            if (!rateQueue.isEmpty()) {
+                // Peek again to confirm it's the one we used.
+                RateChangeQueue.RateChange memory frontChange = rateQueue.peek();
```
Stebalien, line 558: How would this not be the case?

src/Payments.sol
aarshkshah1992: We need to remove this assert.

src/Payments.sol
aarshkshah1992: if the queue is non-empty AND next element has expired -> until epoch is in the past or < targetEpoch -> pop.

src/Payments.sol
```diff
@@ -0,0 +1,612 @@
...
+
+            // If there's an upcoming rate change, check if it applies within our current range.
+            if (!rateQueue.isEmpty()) {
+                RateChangeQueue.RateChange memory upcomingChange = rateQueue.peek();
+                bool isWithinRange = (
+                    upcomingChange.untilEpoch >= currentEpoch &&
+                    upcomingChange.untilEpoch <= targetEpoch
+                );
+                require(isWithinRange, "rate queue is in an invalid state");
+                nextBoundary = upcomingChange.untilEpoch;
```
aarshkshah1992, line 529: min(next, target)

aarshkshah1992: nextBoundary = min(nextBoundary, target)

src/Payments.sol
aarshkshah1992: actually -> we can JUST dequeue if we reach here.

src/Payments.sol
```diff
@@ -0,0 +1,612 @@
...
+                activeRate = upcomingChange.rate;
+            }
+
+            // Settle the segment from the current epoch up to the next boundary (or early if arbitration says so).
+            (uint256 segmentAmount, uint256 segmentEndEpoch, string memory arbNote) = _settleSegment(
+                railId,
+                rail,
+                currentEpoch,
+                nextBoundary,
```
Stebalien, line 538: pass `min(nextBoundary, targetEpoch)`

src/Payments.sol
```diff
@@ -0,0 +1,612 @@
...
+            note = arbResult.note;
+        }
+
+        // Revert if no progress was made (i.e., the final epoch did not advance).
+        require(settledUntil > epochStart, "failed to settle: no progress made in settlement");
+
+        return (settledAmount, settledUntil, note);
+    }
+
+    function _settleWithRateChanges(
```
aarshkshah1992, line 505: mark the function as `nonReEntrant`.

aarshkshah1992: Infact all functions that modify rail state should have a per-rail lock.

aarshkshah1992: Audit and REALLY think through it.

src/Payments.sol
```diff
@@ -0,0 +1,612 @@
...
+     * @param railId The identifier of the rail being settled.
+     * @param rail A storage reference to the rail.
+     * @param epochStart The starting epoch of this settlement segment.
+     * @param epochEnd The ending epoch of this settlement segment.
+     * @param rate The rate for this settlement segment.
+     * @return settledAmount The final amount to settle for this segment.
+     * @return settledUntil The final epoch that effectively got settled.
+     * @return note An optional note returned by the arbitrator.
+     */
+    function _settleSegment(
```
aarshkshah1992, line 454: rentrancy

aarshkshah1992: per-rail settlement lock

src/Payments.sol
```diff
@@ -0,0 +1,612 @@
...
+
+    struct OperatorApproval {
+        bool isApproved;
+        address arbitrer;
+        uint256 rateAllowance;
+        uint256 lockupAllowance;
+    }
+
+    // railId => RateChangeQueue
+    mapping(uint256 => RateChangeQueue.Queue) public railRateChangeQueues;
```
aarshkshah1992, line 61: move this to the Rail struct.

src/Payments.sol
Stebalien: Also, if _operators_ create rails, it makes sense to:

1. Approve the operator with no allowance.
2. Let the operator create rails.
3. Then give them allowance.

(IMO)

Stebalien:
The re-entency issue:

We discussed a few approaches:

1. Lock everything preventing both read and write access to the entire contract while settling a rail. This would restrict access to important information, e.g., information about the rail.
2. Lock the rail only. Unfortunately, if we lock read access, we'd still be restricting information about the rail. Furthermore, if we don't adjust payee/payer funds while settling and still allow, e.g., the arbiter to _see_ those fields, the arbiter won't see the "full" effects of the last section of payment it arbitrated.

We settled on:

1. Locking modifications (preventing settling, changing lockup, etc.) to the rail while settling.
2. Updating the rail, the payee, and the payer after _each_ call to the arbiter such that the arbiter always sees the payment contract in the correct state. Effectively, we move most of the rail settlement logic into the "settle segment" function.

Thinking on this more, there _are_ a few more solutions (which I don't like but I figured I'd mention for completeness).

1. We _could_ not lock anything, record important state before calling the arbiter, call the arbiter, then revert if "something has changed". This "works" but is strictly more brittle and doesn't provide any benefit.
2. We _could_ allow payment/lockup changes as long as we're really careful as those changes can only affect future time periods. But I don't think it's worth it and re-entrant modifications are likely always bugs.

Stebalien:
We also discussed how this relates to withdrawal. IMO, the difference between this and withdrawal is that all the "withdraw" call can do is revert. It can't cause us to change our minds and withdraw _less_ than we initially intended so there are no funky intermediate states.

src/Payments.sol
aarshkshah1992: We agree that we shouldn't have this here as the arbitrer might be different for each rail.

src/Payments.sol
aarshkshah1992: Have changed `createRail` to automagically approve an operator with zero allowance if it does not already exist.

src/Payments.sol

src/Payments.sol
aarshkshah1992: Done.

@Stebalien -> Not only do they not need funds now, but we've fixed this so that both `from` and `to` accounts will be created if they don't already exist.

src/Payments.sol
aarshkshah1992: We agreed that only operators can create rails.

src/Payments.sol
aarshkshah1992: Removed this function.

src/Payments.sol
aarshkshah1992: Done. Removed this.

src/Payments.sol
aarshkshah1992: Have refactored this entire flow to make it easy to reason about.

src/Payments.sol
aarshkshah1992: This is done.

src/Payments.sol
aarshkshah1992: Removed  this. Will add it back if needed.

src/Payments.sol

src/Payments.sol

src/Payments.sol

src/Payments.sol

src/Payments.sol

src/Payments.sol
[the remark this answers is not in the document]

src/Payments.sol

src/Payments.sol
```diff
@@ -0,0 +1,793 @@
...
+                return (0, note);
+            }
+
+            // Ensure arbiter doesn't allow more payment than the maximum possible
+            // for the epochs they're confirming
+            uint256 maxAllowedAmount = rate * (settledUntilEpoch - epochStart);
+            require(
+                result.modifiedAmount <= maxAllowedAmount,
+                "arbiter modified amount exceeds maximum for settled duration"
+            );
```
Stebalien, line 713: I'd:

1. Require that `settledUntilEpoch >= epochStart`.
2. Require that the amount settled is within expectations given the `settledUntilEpoch`.
3. Finally, short-circuit if no funds were actually settled.  Or, TBH, I'd just go to the end of the function anyways and not short-circuit.

Specifically, I don't want to return `(0, note)` if the arbiter returns nonsensical results (e.g., settle 1M FIL up till some epoch in the past).

Stebalien:
This is looking really good. I'd say punt operator termination stuff to a new PR, finish up the last nits here, then we can merge it.

src/Payments.sol
```diff
@@ -0,0 +1,793 @@
...
+
+    modifier noRailModificationInProgress(uint256 railId) {
+        require(
+            !_railModificationLocks[railId],
+            "Modification already in progress"
+        );
+        _railModificationLocks[railId] = true;
+        _;
+        _railModificationLocks[railId] = false;
+    }
```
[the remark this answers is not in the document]
aarshkshah1992: Done.

src/Payments.sol
aarshkshah1992: Added license and made the changes.

src/Payments.sol
```diff
@@ -0,0 +1,793 @@
...
+
+            // Settle the rail up to the current block
+            (, uint256 settledUntilEpoch, ) = settleRail(
+                railIds[i],
+                block.number
+            );
+            require(
+                settledUntilEpoch == block.number,
+                "Failed to settle rail completely"
+            );
```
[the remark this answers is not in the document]

src/Payments.sol
```diff
@@ -0,0 +1,793 @@
...
+
+            // Ensure invariant: lockup should never exceed funds
+            require(
+                account.lockupCurrent <= account.funds,
+                "Lockup exceeds funds after terminating operator"
+            );
+        }
+    }
+
+    function deposit(address token, address to, uint256 amount) external {
```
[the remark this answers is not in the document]

src/Payments.sol
```diff
@@ -0,0 +1,793 @@
...
+    }
+
+    function modifyRailPayment(
+        uint256 railId,
+        uint256 newRate,
+        uint256 oneTimePayment
+    )
+        external
+        validateRailActive(railId)
+        validateRailAccountsExist(railId)
```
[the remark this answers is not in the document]
aarshkshah1992: I don't think we need the ` validateRailAccountsExist` modifier at all now as all accounts always exist implicitly with zero funds (we've now removed the `ownerAddress` field from Account).

src/Payments.sol
```diff
@@ -0,0 +1,793 @@
...
+    }
+
+    function modifyRailLockup(
+        uint256 railId,
+        uint256 period,
+        uint256 lockupFixed
+    )
+        external
+        validateRailActive(railId)
+        validateRailAccountsExist(railId)
```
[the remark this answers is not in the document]
aarshkshah1992: I don't think we need the  validateRailAccountsExist modifier at all now as all accounts always exist implicitly with zero funds (we've now removed the ownerAddress field from Account).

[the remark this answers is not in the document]
aarshkshah1992: Done.

src/Payments.sol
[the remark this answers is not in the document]
aarshkshah1992: Done.

src/Payments.sol
```diff
@@ -0,0 +1,793 @@
...
+
+    function deposit(address token, address to, uint256 amount) external {
+        require(token != address(0), "token address cannot be zero");
+        require(to != address(0), "to address cannot be zero");
+        require(amount > 0, "amount must be greater than 0");
+
+        // Create account if it doesn't exist
+        Account storage account = accounts[token][to];
+        if (account.ownerAddress == address(0)) {
+            account.ownerAddress = to;
```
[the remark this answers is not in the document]
aarshkshah1992: Yeah, you are right. Removing this field and also removing the `onlyAccount` modifier simplifies things a bit and we can always infer the account owner as being the msg sender for any methods that need the sender to prove account ownership.

src/Payments.sol
```diff
@@ -0,0 +1,694 @@
...
+        IERC20(token).safeTransferFrom(msg.sender, address(this), amount);
+
+        // Update account balance
+        account.funds += amount;
+
+        // settle account lockup now that we have more funds
+        settleAccountLockup(account);
+    }
+
+    function withdraw(address token, uint256 amount) external nonReentrant {
```
aarshkshah1992, line 159: @Stebalien 

I have also implemented a `function withdrawTo(
        address token,
        address to,
        uint256 amount
    )`
    
    function that allows users to withdraw funds to any account of their choosing.

src/Payments.sol
```diff
@@ -0,0 +1,793 @@
...
+
+        // Check if operator is approved, if not, auto-approve with zero allowances
+        OperatorApproval storage approval = operatorApprovals[token][from][
+            operator
+        ];
+        if (!approval.isApproved) {
+            approval.isApproved = true;
+            approval.rateAllowance = 0;
+            approval.lockupAllowance = 0;
+        }
```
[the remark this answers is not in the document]

src/Payments.sol
```diff
@@ -0,0 +1,793 @@
...
+        onlyRailOperator(railId)
+        noRailModificationInProgress(railId)
+    {
+        Rail storage rail = rails[railId];
+        Account storage payer = accounts[rail.token][rail.from];
+        OperatorApproval storage approval = operatorApprovals[rail.token][
+            rail.from
+        ][rail.operator];
+
+        require(approval.isApproved, "operator not approved");
```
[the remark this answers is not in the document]
aarshkshah1992: @Stebalien Let's talk about this in our 1:1 today.

src/Payments.sol
```diff
@@ -0,0 +1,793 @@
...
+        // --- Settlement Prior to Rate Change ---
+        // If there is no arbiter, settle the rail immediately.
+        if (rail.arbiter == address(0)) {
+            (, uint256 settledUpto, ) = settleRail(railId, block.number);
+            require(
+                settledUpto == block.number,
+                "not able to settle rail upto current epoch"
+            );
+        } else {
+            rail.rateChangeQueue.enqueue(rail.paymentRate, block.number);
```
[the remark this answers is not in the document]
aarshkshah1992: @Stebalien Good catch ! How do we want to implement this ?

If a user calls `modifyRailPayment` multiple times in a single epoch -> should we only enque the previous (before this epoch) `rail.paymentRate ` ONLY once and ensure that `rail.paymentRate ` reflects the last rate update ?

Basically, the rate changes within a single epoch do not end up on the queue.

src/Payments.sol
aarshkshah1992: This is done.

aarshkshah1992: TODO for @aarshkshah1992 : Audit the code for off by ones to enforce that the `untilEpoch` in the rate-change queue is not inclusive of the end epoch and that `rail.settledUpto` is inclusive of the settledEpoch.

src/Payments.sol
```diff
@@ -0,0 +1,793 @@
...
+        // Update account balance
+        account.funds += amount;
+
+        // settle account lockup now that we have more funds
+        settleAccountLockup(account);
+    }
+
+    function withdraw(
+        address token,
+        uint256 amount
```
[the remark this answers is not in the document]
aarshkshah1992: @Stebalien Implemented this.

src/Payments.sol
aarshkshah1992: Will do this in the follow-up PR.

aarshkshah1992:
@Stebalien 

I've addressed all your comments.

Only one comment remains unresolved at https://github.com/FilOzone/payments/pull/1#discussion_r1976817452. Let's go over it in our sync today and we can then merge it.

I am now starting work on raising a new PR to implement operator termination and all the other design changes we've added to  https://www.notion.so/filecoindev/Payments-Design-Iteration-17ddc41950c180468806c7e0097c2e14?pvs=4 since our first draft.

aarshkshah1992:
@Stebalien Fixed the CI and removed the `configure` script.

src/Payments.sol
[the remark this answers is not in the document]
aarshkshah1992: @Stebalien I also realised after looking at this again is that we're basically saying that the new rate will be applicable AFTER the current block and rail will be settled upto and including this epoch using the old rate. 

Just wanted to confirm if that's the expected behaviour. IMO, it makes sense.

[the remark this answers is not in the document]
Stebalien: > Basically, the rate changes within a single epoch do not end up on the queue.

Exactly.

> Just wanted to confirm if that's the expected behaviour. IMO, it makes sense.

IMO we could do it either way but... I'd keep it the current way as it's simpler.

src/RateChangeQueue.sol
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    struct Queue {
+        // Map from index to RateChange
+        mapping(uint256 => RateChange) changes;
+        uint256 head;
+        uint256 tail;
+    }
+
+    function enqueue(Queue storage queue, uint256 rate, uint256 untilEpoch) internal {
```
Stebalien, line 17: TODO: drop enqueued rate-changes if `queue.changes[queue.tail].untilEpoch == untilEpoch`.

src/Payments.sol
```diff
@@ -0,0 +1,700 @@
...
+        // --- Settlement Prior to Rate Change ---
+        // If there is no arbiter, settle the rail immediately.
+        if (rail.arbiter == address(0)) {
+            (, uint256 settledUpto, ) = settleRail(railId, block.number);
+            require(
+                settledUpto == block.number,
+                "not able to settle rail upto current epoch"
+            );
+        } else {
+            rail.rateChangeQueue.enqueue(rail.paymentRate, block.number);
```
Stebalien, line 329: We should probably skip all of this if the rate hasn't changed at all (if we're _just_ making a one-time payment. But it should be a no-op otherwise so I'm happy doing this in a new PR.

Stebalien (approved):
LGTM. There are some optimizations we can make, but they aren't required for correctness.

Stebalien:
I'll let you press the merge button when you wake up.

src/Payments.sol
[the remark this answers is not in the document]
aarshkshah1992: This is done.

src/RateChangeQueue.sol
aarshkshah1992: @Stebalien I want to keep the queue implementation simpler. I've made this change at the call site in `modifyRailPayment`.

src/Payments.sol
aarshkshah1992: Doing this in the next PR.

src/Payments.sol
```diff
@@ -326,7 +326,14 @@ contract Payments is
                 "not able to settle rail upto current epoch"
             );
         } else {
-            rail.rateChangeQueue.enqueue(rail.paymentRate, block.number);
+            // handle multiple rate changes by the operator in a single epoch
+            // by queueuing the previous rate only once
+            if (
+                rail.rateChangeQueue.isEmpty() ||
```
aarshkshah1992, line 332: @Stebalien FYI

PR #2, opened 2025-03-03
Add docs to all settlement and lockup epochs 

aarshkshah1992:
Added docs to make it easy to reason about the absence of off by one errors in settlement, lockup and arbitration.


Stebalien (approved):

aarshkshah1992:
@Stebalien Wait, what do you mean ? They are actually [start, stop) ....

PR #3, opened 2025-03-04
feat: enable rail terminate and allow operator to reduce rail rate if rail is not in debt

aarshkshah1992:
- Enable client to terminate rail if rail is not in debt
- Allow operator to reduce rail rate if rail is not in debt and client has enough locked funds to settle the rail upto and including the current epoch.

### DIFF
--- a/src/Payments.sol
+++ b/src/Payments.sol
@@ -591,8 +591,7 @@ contract Payments is
     ) internal view {
         if (isRailTerminated(rail)) {
             require(
-                block.number <=
-                    maxSettlementEpochForTerminatedRail(rail),
+                block.number <= maxSettlementEpochForTerminatedRail(rail),
                 "terminated rail is beyond it's max settlement epoch; settle the rail"
             );
             require(
@@ -832,8 +831,7 @@ contract Payments is
         // Check if rail is a terminated rail that's now fully settled
         if (
             isRailTerminated(rail) &&
-            rail.settledUpTo >=
-            maxSettlementEpochForTerminatedRail(rail)
+            rail.settledUpTo >= maxSettlementEpochForTerminatedRail(rail)
         ) {
             finalizeTerminatedRail(rail, payer);
             return (amount, finalEpoch, finalizedNote);
--- a/src/Payments.sol
+++ b/src/Payments.sol
@@ -463,6 +463,19 @@ contract Payments is
             "one time payment cannot be greater than rail lockupFixed"
         );
 
+        // --- Operator Approval Checks
+        OperatorApproval storage approval = operatorApprovals[rail.token][
+            rail.from
+        ][rail.operator];
+        validateAndModifyRateChangeApproval(
+            rail,
+            approval,
+            oldRate,
+            newRate,
+            oneTimePayment,
+            effectiveLockupPeriod
+        );
+
         // Update the rail fixed lockup and payment rate
         rail.lockupFixed = rail.lockupFixed - oneTimePayment;
         rail.paymentRate = newRate;
@@ -484,19 +497,6 @@ contract Payments is
             (oldRate * effectiveLockupPeriod) +
             (newRate * effectiveLockupPeriod) -
             oneTimePayment;
-
-        // --- Operator Approval Checks AFTER calculating the new lockups and rates ---
-        OperatorApproval storage approval = operatorApprovals[rail.token][
-            rail.from
-        ][rail.operator];
-        validateAndModifyRateChangeApproval(
-            rail,
-            approval,
-            oldRate,
-            newRate,
-            oneTimePayment,
-            effectiveLockupPeriod
-        );
 
         // --- Process the One-Time Payment ---
         processOneTimePayment(payer, payee, oneTimePayment);
@@ -664,16 +664,19 @@ contract Payments is
         uint256 oldLockup = (oldRate * effectiveLockupPeriod) +
             rail.lockupFixed;
         uint256 newLockup = (newRate * effectiveLockupPeriod) +
-            rail.lockupFixed;
+            (rail.lockupFixed - oneTimePayment);
 
-        // note: we've reduced `rail.lockupFixed` by `oneTimePayment` before calling this
-        // so we need to add the oneTimePayment to the new lockup for allowance tracking
-        // as all one time payments are also bound by the operator lockup allowance
-        updateOperatorLockupTracking(
-            approval,
-            oldLockup,
-            newLockup + oneTimePayment
-        );
+        updateOperatorLockupTracking(approval, oldLockup, newLockup);
+
+        if (oneTimePayment > 0) {
+            // one-time payments count towards lockup usage
+            require(
+                approval.lockupUsage + oneTimePayment <=
+                    approval.lockupAllowance,
+                "one-time payment exceeds operator lockup allowance"
+            );
+            approval.lockupUsage += oneTimePayment;
+        }
 
         // handle a rate decrease
         if (newRate < oldRate) {
--- a/src/Payments.sol
+++ b/src/Payments.sol
@@ -166,20 +166,11 @@ contract Payments is
         approval.isApproved = true;
     }
 
-    function setOperatorApprovalStatus(
+
+    function setOperatorApproval(
         address token,
         address operator,
-        bool approved
-    ) external {
-        require(token != address(0), "token address cannot be zero");
-        require(operator != address(0), "operator address cannot be zero");
-
-        operatorApprovals[token][msg.sender][operator].isApproved = approved;
-    }
-
-    function setOperatorAllowances(
-        address token,
-        address operator,
+        bool    approved,
         uint256 rateAllowance,
         uint256 lockupAllowance
     ) external {
@@ -189,10 +180,13 @@ contract Payments is
         OperatorApproval storage approval = operatorApprovals[token][
             msg.sender
         ][operator];
+        
+        // Update approval status and allowances
+        approval.isApproved = approved;
         approval.rateAllowance = rateAllowance;
         approval.lockupAllowance = lockupAllowance;
     }
-
+    
     function terminateRail(
         uint256 railId
     ) external validateRailActive(railId) noRailModificationInProgress(railId) {
--- a/src/Payments.sol
+++ b/src/Payments.sol
@@ -111,6 +111,14 @@ contract Payments is
         rails[railId].isLocked = false;
     }
 
+    modifier onlyRailClient(uint256 railId) {
+        require(
+            rails[railId].from == msg.sender,
+            "only the rail client can perform this action"
+        );
+        _;
+    }
+
     modifier onlyRailOperator(uint256 railId) {
         require(
             rails[railId].operator == msg.sender,
@@ -225,40 +233,35 @@ contract Payments is
 
     function terminateRail(
         uint256 railId
-    ) external validateRailActive(railId) noRailModificationInProgress(railId) {
+    )
+        external
+        validateRailActive(railId)
+        onlyRailClient(railId)
+        noRailModificationInProgress(railId)
+    {
         Rail storage rail = rails[railId];
 
-        // Ensure only the payer can terminate the rail
-        require(
-            rail.from == msg.sender,
-            "only the rail payer can terminate the rail"
-        );
-
         Account storage payer = accounts[rail.token][rail.from];
-
-        // Settle account lockup
-        settleAccountLockup(payer);
-
-        // Check that the rail is not in debt
-        // (i.e., client has enough funds to pay for services already taken on this rail upto and including the current epoch)
+        // Settle account lockup and ensure it's fully settled up to the current block
+        (bool fullySettled, uint256 settledEpoch) = settleAccountLockup(payer);
         require(
-            block.number < payer.lockupLastSettledAt + rail.lockupPeriod,
-            "rail is in debt; cannot terminate"
+            fullySettled && settledEpoch == block.number,
+            "account lockup must be fully settled to terminate rail"
         );
 
-        // Calculate the effective remaining lockup period that hasn't been settled
-        require(
-            payer.lockupLastSettledAt <= block.number,
-            "lockup settlement epoch cannot be in the future"
-        );
-        uint256 effectiveLockupPeriod = rail.lockupPeriod -
-            (block.number - payer.lockupLastSettledAt);
-
-        // Calculate funds that can be released (funds locked for future epochs + fixed lockup)
-        // note: this does not include the current epoch as the rail still needs to be paid for for the
-        // current epoch
-        uint256 fundsToRelease = (rail.paymentRate * effectiveLockupPeriod) +
-            rail.lockupFixed;
+        // If there is no arbiter, settle the rail immediately and ensure it is fully settled upto and including
+        // the current epoch
+        if (rail.arbiter == address(0)) {
+            (, uint256 settledUpto, ) = settleRail(railId, block.number);
+            require(
+                settledUpto == block.number,
+                "rail must be fully settled to terminate"
+            );
+        } else {
+            // For arbitrated rails, push the current rate to the queue
+            // This ensures proper settlement when the operator later settles this rail as we're setting the current rail rate to zero in this function
+            rail.rateChangeQueue.enqueue(rail.paymentRate, block.number);
+        }
 
         // Update payer's lockup rate and current lockup
         require(
@@ -267,18 +270,8 @@ contract Payments is
         );
         payer.lockupRate = payer.lockupRate - rail.paymentRate;
 
-        require(
-            payer.lockupCurrent >= fundsToRelease,
-            "payer lockup current must be at least funds to release"
-        );
-        payer.lockupCurrent = payer.lockupCurrent - fundsToRelease;
-
-        // Set rail payment rate and lockup period to 0 for future periods
+        // Set rail payment rate to 0 for future periods and mark the rail as inactive
         rail.paymentRate = 0;
-        rail.lockupPeriod = 0;
-        rail.lockupFixed = 0;
-
-        // Mark rail as inactive
         rail.isActive = false;
 
         // Ensure account invariants hold
@@ -542,7 +535,6 @@ contract Payments is
         uint256 untilEpoch
     )
         public
-        validateRailActive(railId)
         nonReentrant
         returns (
             uint256 totalSettledAmount,
--- a/src/Payments.sol
+++ b/src/Payments.sol
@@ -256,8 +256,8 @@ contract Payments is
 
         Account storage acct = accounts[token][msg.sender];
 
-        (bool funded, uint256 settleEpoch) = settleAccountLockup(acct);
-        require(funded && settleEpoch == block.number, "insufficient funds");
+        uint256 settleEpoch = settleAccountLockup(acct);
+        require(settleEpoch == block.number, "insufficient funds");
 
         uint256 available = acct.funds > acct.lockupCurrent
             ? acct.funds - acct.lockupCurrent
@@ -329,11 +329,9 @@ contract Payments is
         ][rail.operator];
 
         // settle account lockup and if account lockup is not settled upto to the current epoch; revert
-        (bool fullySettled, uint256 lockupSettledUpto) = settleAccountLockup(
-            payer
-        );
+        uint256 lockupSettledUpto = settleAccountLockup(payer);
         require(
-            fullySettled && lockupSettledUpto == block.number,
+            lockupSettledUpto == block.number,
             "cannot modify lockup as client does not have enough funds for current account lockup"
         );
 
@@ -403,14 +401,10 @@ contract Payments is
         ][rail.operator];
 
         // Settle the payer's lockup to account for elapsed time
-        (bool fullySettled, uint256 lockupSettledUpto) = settleAccountLockup(
-            payer
-        );
-
+        uint256 lockupSettledUpto = settleAccountLockup(payer);
         // If we're increasing the rate, we require full account lockup setlement before we can proceed
         require(
-            newRate <= oldRate ||
-                (fullySettled && lockupSettledUpto == block.number),
+            newRate <= oldRate || lockupSettledUpto == block.number,
             "account lockup not fully settled; cannot increase rate"
         );
 
@@ -518,11 +512,8 @@ contract Payments is
         // If we've reduced the rate, settle lockup again to account for changes
         // and ensure that account lockup is settled upto and including the current epoch
         if (newRate < oldRate) {
-            (bool settled, uint256 settledUpto) = settleAccountLockup(payer);
-            require(
-                settled && settledUpto == block.number,
-                "account lockup must be fully settled after rate decrease"
-            );
+            uint256 settledUpto = settleAccountLockup(payer);
+            require(settledUpto == block.number, "account lockup must be fully settled after rate decrease");
         }
     }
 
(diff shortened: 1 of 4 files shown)


src/Payments.sol
```diff
@@ -223,6 +223,71 @@ contract Payments is
...
+            "lockup settlement epoch cannot be in the future"
+        );
+        uint256 effectiveLockupPeriod = rail.lockupPeriod -
+            (block.number - payer.lockupLastSettledAt);
+
+        // Calculate funds that can be released (funds locked for future epochs + fixed lockup)
+        // note: this does not include the current epoch as the rail still needs to be paid for for the
+        // current epoch
+        uint256 fundsToRelease = (rail.paymentRate * effectiveLockupPeriod) +
+            rail.lockupFixed;
```
Stebalien, line 261: I want to avoid locking new funds but, IMO, terminating rails this way shouldn't release locked funds. Terminate is really "SIGKILL". The "nice" way to stop a rail is to terminate all deals on that rail through the rails operator. But, if the rail is buggy, this lets the client say "ok, no more".

src/Payments.sol
```diff
@@ -324,58 +401,89 @@ contract Payments is
...
-                rail.rateChangeQueue.peek().untilEpoch != block.number
-            ) {
-                // For arbitrated rails, we need to enqueue the old rate.
-                // This ensures that the old rate is applied up to and including the current block.
-                // The new rate will be applicable starting from the next block.
-                rail.rateChangeQueue.enqueue(rail.paymentRate, block.number);
+        // Only settle and update rate queue if the rate has changed
+        if (newRate != oldRate) {
+            // If there is no arbiter, settle the rail immediately.
+            if (rail.arbiter == address(0)) {
```
Stebalien, line 407: code nit:

```sol
if (newRate == oldRate) {
    // do nothing
} else if ...
```

Stebalien (commented):
Note: I rebased this against main.

aarshkshah1992:
@Stebalien Thoughts on operator terminstion   ?

src/Payments.sol
aarshkshah1992: @Stebalien I see. So on rail termination, we ideally want to keep the future(rail period) funds locked but allows a client to exit the rail _after_ that.

src/Payments.sol
```diff
@@ -223,6 +223,71 @@ contract Payments is
...
+
+        // Settle account lockup
+        settleAccountLockup(payer);
+
+        // Check that the rail is not in debt
+        // (i.e., client has enough funds to pay for services already taken on this rail upto and including the current epoch)
+        require(
+            block.number < payer.lockupLastSettledAt + rail.lockupPeriod,
+            "rail is in debt; cannot terminate"
+        );
```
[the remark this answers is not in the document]
aarshkshah1992: I mean sure but this atleast assumes the client wants to do the right thing. 

If the client does not cover their debt -> there's nothing we can do but we should now allow them to exit their on-chain obligations and keep a record of that around.

What does allowing rail termination of an indebted rail get the client or the operator ?

src/Payments.sol
Stebalien: Less indentation. It doesn't really matter, I just try to avoid too much indentation where possible.

src/Payments.sol
aarshkshah1992: @Stebalien We need to discuss this a bit. If we set the rate to zero here, rail settlement wont actually settle anything.

src/Payments.sol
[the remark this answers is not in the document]
Stebalien: > What does allowing rail termination of an indebted rail get the client or the operator ?

I was approaching this from a different perspective:

1. Could a client "terminate" a rail some other way? The answer is yes: they use separate accounts per rail and just stop funding individual accounts.
2. Is this a problem for the payment system? Kind of. It means that a rational client will jump through a bunch of hoops to avoid re-using the same account for multiple payments just in case.

On the other hand, assuming the client has a non-zero lockup period, they had plenty of time to terminate the rail before they went into debt. But that also means that the SP had plenty of warning that the client was likely going to go into debt and they should have started canceling client services already.

So... I'm not really sure.

src/Payments.sol
Stebalien: Yes. It cuts off the flow of funds to the rail but still preserves all the other guarantees. Otherwise, locking funds is kind of pointless.

An alternative (potential addition?) to terminate is "freeze": the ability to freeze all funds into a rail while still allowing funds to flow out. Eventually the rail would go into debt. I bring this up because it _might_ be useful in some cases where there are disputes that need to be handled out-of-band. To "unfreeze" a rail, the client would have to fully fund it. But, IMO, that's a separate feature we should consider later.

Stebalien:
> @Stebalien Thoughts on operator termination?

After thinking about how this plays into rail termination, I wonder if we do just want to use the allowance. Basically:

1. The client _can_ lower the operator's allowance below the sum of the lockup rates on all rails managed by that operator. This doesn't affect the current rails, but it means that the operator cannot increase rates on any rails until they drop below their new allowance.
2. To "terminate" an operator, the client would set the operator's allowance to 0, then terminate each rail managed by the operator to stop the flow of funds.

src/Payments.sol
aarshkshah1992: @Stebalien I have refactored this to ensure both account lockup and non-arbitrated rails are fully settled before this can happen. For arbitrated rails -> we push the current rate to the queue.

We then set rail rate to 0.

Please take a look.

aarshkshah1992:
Client should be able to modify an existing allowance.

src/Payments.sol
```diff
@@ -223,6 +231,56 @@ contract Payments is
...
+
+        // Update payer's lockup rate and current lockup
+        require(
+            payer.lockupRate >= rail.paymentRate,
+            "payer lockup rate must be at least rail payment rate"
+        );
+        payer.lockupRate = payer.lockupRate - rail.paymentRate;
+
+        // Set rail payment rate to 0 for future periods and mark the rail as inactive
+        rail.paymentRate = 0;
```
Stebalien, line 274: We don't want to lock _new_ funds, but we still need to allow paying out the already locked funds up till the end of the lockup period.

src/Payments.sol
```diff
@@ -134,11 +169,74 @@ contract Payments is
...
+
+        operatorApprovals[token][msg.sender][operator].isApproved = approved;
+    }
+
+    function setOperatorAllowances(
+        address token,
+        address operator,
+        uint256 rateAllowance,
+        uint256 lockupAllowance
+    ) external {
```
Stebalien, line 203: Maybe just have a single function that does both at the same time?

src/Payments.sol
```diff
@@ -119,6 +134,26 @@ contract Payments is
         _;
     }
 
+    function terminateOperator(address operator) external {
```
Stebalien, line 137: Hm. Looking at how this turned out.... I'm wondering if we want to just do this off-chain. I.e., make the client send one message per operator per token to unapprove and set the allowances to zero. Up to you, I'm just not sure if the extra complexity/state is worth it.

src/Payments.sol
```diff
@@ -296,24 +418,47 @@ contract Payments is
...
+        // For terminated rails, can only reduce rate, not increase it
+        if (rail.terminationEpoch > 0) {
+            require(
+                newRate <= oldRate,
+                "failed because terminated at epoch: cannot increase rate on terminated rail"
+            );
+            require(
+                oneTimePayment <= rail.lockupFixed,
+                "failed because terminated at epoch: one-time payment exceeds fixed lockup"
+            );
```
Stebalien, line 432: This is already required in all cases, right?

Stebalien (commented):
still need to go through the rest but my brain is fried for today.

src/Payments.sol
```diff
@@ -296,24 +424,47 @@ contract Payments is
...
+        );
+
+        // We just settled the account lockup. If the client is still not able to guaruantee
+        // enough funds to lock up for the rail upto and including the current epoch, it means that the rail is in "debt"
+        // i.e. the client does not have enough funds to pay for services already taken on this rail.
+        // In this case, we should not allow rate modifications at all.
+        // The client should first deposit enough funds to pay for this "debt" and then the rate can be changed.
+        require(
+            newRate == oldRate ||
+                block.number < payer.lockupLastSettledAt + rail.lockupPeriod,
```
Stebalien, line 464:
```suggestion
                block.number <= payer.lockupLastSettledAt + rail.lockupPeriod,
```

I think?

src/Payments.sol
```diff
@@ -296,24 +424,47 @@ contract Payments is
...
-                fullySettled && lockupSettledUpto == block.number,
-                "account lockup not fully settled; cannot change rate"
-            );
-        }
 
-        uint256 oldRate = rail.paymentRate;
+        // If we're increasing the rate, we require full account lockup setlement before we can proceed
+        require(
+            newRate <= oldRate ||
+                (fullySettled && lockupSettledUpto == block.number),
```
Stebalien, line 453: Can it be `fullySettled` and not `lockupSettledUpto == block.number`?

src/Payments.sol
```diff
@@ -296,24 +424,47 @@ contract Payments is
...
+            "account lockup not fully settled; cannot increase rate"
+        );
+
+        // We just settled the account lockup. If the client is still not able to guaruantee
```
Stebalien, line 457: I'd find this easier to understand if we broke it down case by case as separate conditionals:

1. Case 1: lockup not fully settled.
    1. Case 1a: in debt: require that rate must not change.
    2. Case 1b: not in debt: require that rate is less than or equal to the current rate.
3. Case 2: lockup fully settled, nothing to check.

src/Payments.sol
```diff
@@ -324,58 +475,95 @@ contract Payments is
...
+                require(
+                    settledUpto == block.number,
+                    "not able to settle rail upto current epoch"
+                );
+            } else {
+                // handle multiple rate changes by the operator in a single epoch
+                // by queueuing the previous rate only once
+                if (
+                    rail.rateChangeQueue.isEmpty() ||
+                    rail.rateChangeQueue.peek().untilEpoch != block.number
```
Stebalien, line 496: Gah. This is the head, we need to look at the tail. Didn't catch this in the original PR.

src/Payments.sol
```diff
@@ -324,58 +475,95 @@ contract Payments is
...
+                    rail.rateChangeQueue.enqueue(oldRate, block.number);
+                }
             }
         }
 
+        // Calculate the effective lockup period (remaining period that hasn't been settled)
+        require(
+            payer.lockupLastSettledAt <= block.number,
+            "lockup settlement epoch cannot be in the future"
+        );
```
Stebalien, line 510: Eh... we've already asserted this when we settled the lockup right?

Stebalien: (again, note that solidity does check for underflow/overflow)

src/Payments.sol
```diff
@@ -324,58 +475,95 @@ contract Payments is
...
+        uint256 effectiveLockupPeriod = rail.lockupPeriod -
+            (block.number - payer.lockupLastSettledAt);
```
Stebalien, line 512: We're going to have to take the termination epoch into account here. I.e. `min(payer.lockupLastSettledAt, rail.terminationEpoch)` (if `rail.terminationEpoch > 0`). I think it's time for some helper functions.

Stebalien (commented):
Another partial review. We're reaching a complexity breaking point. I think we can partially solve this with some careful helper functions, but I'll have to think about it a bit.

src/Payments.sol
aarshkshah1992: Done.

src/Payments.sol
aarshkshah1992: Done. We now have a single `setOperatorApproval` that subsumes both the `setOperatorAllowances` AND `setOperatorApprovalStatus` functions here.

src/Payments.sol
```diff
@@ -134,11 +175,74 @@ contract Payments is
...
+    ) external validateRailActive(railId) noRailModificationInProgress(railId) {
+        Rail storage rail = rails[railId];
+
+        // Only the client, operator, or payee can terminate the rail
+        require(
+            msg.sender == rail.from ||
+                msg.sender == rail.operator ||
+                msg.sender == rail.to,
+            "caller cannot terminate this rail"
+        );
```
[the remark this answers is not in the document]
aarshkshah1992: Done.

src/Payments.sol
```diff
@@ -134,11 +175,74 @@ contract Payments is
...
+
+        require(rail.terminationEpoch == 0, "rail already terminated");
```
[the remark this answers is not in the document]
aarshkshah1992: Done.

src/Payments.sol
```diff
@@ -134,11 +175,74 @@ contract Payments is
...
+        // Set termination epoch to current block
+        rail.terminationEpoch = block.number;
+
+        // Remove the rail rate from account lockup rate but don't set rail rate to zero yet
+        Account storage payer = accounts[rail.token][rail.from];
+        require(
+            payer.lockupRate >= rail.paymentRate,
+            "lockup rate inconsistency"
+        );
+        payer.lockupRate -= rail.paymentRate;
```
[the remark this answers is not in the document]
aarshkshah1992: I'd like to leave them not because they are checking for underflow but also because they can help debug some important logical invariants being broken in the code.

src/Payments.sol
aarshkshah1992: Done.

src/Payments.sol
```diff
@@ -255,15 +371,27 @@ contract Payments is
             (rail.paymentRate * rail.lockupPeriod);
         uint256 newLockup = lockupFixed + (rail.paymentRate * period);
```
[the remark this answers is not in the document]
aarshkshah1992: Done.

src/Payments.sol
```diff
@@ -255,15 +371,27 @@ contract Payments is
...
+
+            // Update usage
+            approval.lockupUsage += lockupIncrease;
+        } else if (newLockup < oldLockup) {
+            uint256 lockupDecrease = oldLockup - newLockup;
+
+            // If decreasing, reduce usage (ensuring no underflow)
+            approval.lockupUsage = approval.lockupUsage > lockupDecrease
+                ? approval.lockupUsage - lockupDecrease
+                : 0;
```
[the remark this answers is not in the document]
aarshkshah1992: @Stebalien Please note that we don't check usage against allowance after REDUCING the lockup here anymore as we need to accomodate the scenario where a client reduces operator allowance after a rail has been created.

So, the scenario where an operator is in allowance debt is now a valid one but we SHOULD allow the operator to reduce their lockup in this case.

src/Payments.sol
aarshkshah1992: Also simplified the asserts here.

src/Payments.sol
aarshkshah1992: no this happens. I've simplified the `settleAccountLockup` function to return only one value now.

src/Payments.sol
aarshkshah1992: Ooops, fixed !

src/Payments.sol
```diff
@@ -707,10 +985,87 @@ contract Payments is
             uint256 fractionalEpochs = availableFunds / account.lockupRate;
             settledUpto = account.lockupLastSettledAt + fractionalEpochs;
 
+            //
```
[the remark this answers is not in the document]
aarshkshah1992: Fixed.

src/Payments.sol
aarshkshah1992: Yeah, you are right. a rail is in debt only if it is not able to pay upto and including the current epoch.

src/Payments.sol
aarshkshah1992: Yeah. Fixed.

src/RateChangeQueue.sol
```diff
@@ -39,6 +39,13 @@ library RateChangeQueue {
         require(queue.head < queue.tail, "Queue is empty");
         return queue.changes[queue.head];
     }
+    
+    function peekTail(
+        Queue storage queue
+    ) internal view returns (RateChange memory) {
```
aarshkshah1992, line 45: @Stebalien Added this to fix the bug with repeated rate changes in a single epoch.

src/Payments.sol
```diff
@@ -136,9 +181,44 @@ contract Payments is
...
+        ][operator];
+        
+        // Update approval status and allowances
+        approval.isApproved = approved;
+        approval.rateAllowance = rateAllowance;
+        approval.lockupAllowance = lockupAllowance;
+    }
+
+    function terminateRail(
+        uint256 railId
```
aarshkshah1992, line 206: @Stebalien Do we need to settle lockup upto current epoch here ?

src/Payments.sol
```diff
@@ -296,86 +388,136 @@ contract Payments is
...
+        uint256 standardRemainingPeriod = rail.lockupPeriod - (block.number - payer.lockupLastSettledAt);
+        
+        if (rail.terminationEpoch > 0) {
+            uint256 terminationRemainingPeriod = handleTerminatedRailRateChange(rail);
+            
+            // Take the minimum of standard and termination remaining periods
+            // This ensures we don't over-lockup - we're constrained by both:
+            // 1. How much of the lockup period hasn't been settled
+            // 2. How much time remains until termination is complete
+            effectiveLockupPeriod = min(standardRemainingPeriod, terminationRemainingPeriod);
```
aarshkshah1992, line 425: I think this is incorrect

src/Payments.sol
```diff
@@ -392,49 +534,192 @@ contract Payments is
...
             oneTimePayment <= rail.lockupFixed,
             "one-time payment exceeds rail fixed lockup"
         );
 
-        // Calculate the original total lockup amount:
         uint256 oldTotalLockup = (oldRate * rail.lockupPeriod) +
             rail.lockupFixed;
-        uint256 newTotalLockup = (newRate * rail.lockupPeriod) +
+        uint256 newTotalLockup = (newRate * effectiveLockupPeriod) +
             rail.lockupFixed;
```
aarshkshah1992, line 660: @Stebalien I am not sure how to calculate the old lockup here

src/Payments.sol
```diff
@@ -119,6 +136,34 @@ contract Payments is
...
+        );
+        _;
+    }
+
+    modifier validateRailNotTerminated(uint256 railId) {
+        require(rails[railId].terminationEpoch == 0, "rail already terminated");
+        _;
+    }
+
+    function terminateOperator(address operator, address token) external {
```
Stebalien, line 154: I think we can drop this and leave only `setOperatorApproval`.

src/Payments.sol
Stebalien: Yes. I think we could get away to ensuring that `lockupSettledAt + rail.lockupPeriod >= block.number`, but let's punt on that for now.

src/Payments.sol
```diff
@@ -234,36 +315,47 @@ contract Payments is
...
+        uint256 lockupSettledUpto = settleAccountLockup(payer);
         require(
-            fullySettled && lockupSettledUpto == block.number,
+            lockupSettledUpto == block.number,
             "cannot modify lockup as client does not have enough funds for current account lockup"
         );
 
         // Calculate the change in base lockup
         uint256 oldLockup = rail.lockupFixed +
             (rail.paymentRate * rail.lockupPeriod);
```
Stebalien, line 340: This old lockup is wrong for terminated rails because the period shrinks over time. Instead of `calculateTerminatedRailLockup`, I think we want a `getEffectiveLockupPeriod(rail)` function that returns the effective lockup period of the rail whether or not it has been terminated. Then we can use that same period for both old lockup and new lockup.

Stebalien: Also, we should probably rename these variables to `oldFutureLockup` and `newFutureLockup`. They don't account for lockup that goes towards _existing_ (past) debts.

src/Payments.sol
```diff
@@ -234,36 +315,47 @@ contract Payments is
...
-        require(approval.isApproved, "operator not approved");
-
         // settle account lockup and if account lockup is not settled upto to the current epoch; revert
-        (bool fullySettled, uint256 lockupSettledUpto) = settleAccountLockup(
-            payer
-        );
+        uint256 lockupSettledUpto = settleAccountLockup(payer);
         require(
-            fullySettled && lockupSettledUpto == block.number,
+            lockupSettledUpto == block.number,
```
Stebalien, line 334: note: technically it should be fine if _either_ the rail is terminated or lockup has been settled. Terminated rails don't lock any additional funds.

(we can leave this for now, but we may want to change it in the future)

src/Payments.sol
Stebalien: The calculation should be the same as the new `effectiveLockupPeriod`, just with the old period in case the period changed. That is, it's the period minus the time between the termination epoch and the current epoch (if any).

That assumes lockup was fully up-to-date at termination but... I think we should stick with that for now.

src/Payments.sol

src/Payments.sol
aarshkshah1992: We now ensure that the account lockup is fully settled upto and including the current epoch before a rail is terminated.

src/Payments.sol
aarshkshah1992: We do now ensure that the account lockup is fully settled during rail termination. It's a reasonable ask from the user and makes a whole host of things easy to reason about.

src/Payments.sol
```diff
@@ -393,45 +571,128 @@ contract Payments is
...
-        uint256 newTotalLockup = (newRate * rail.lockupPeriod) +
+        uint256 newLockup = (newRate * effectiveLockupPeriod) +
             rail.lockupFixed;
 
-        require(
-            newTotalLockup <= oldTotalLockup + approval.lockupAllowance,
-            "exceeds operator lockup allowance"
+        // note: we've reduced `rail.lockupFixed` by `oneTimePayment` before calling this
+        // so we need to add the oneTimePayment to the new lockup for allowance tracking
+        // as all one time payments are also bound by the operator lockup allowance
```
Stebalien, line 669: This looks correct, but it's really confusing.

Stebalien: (not sure how to improve it, but we should consider revisiting it in a future PR)

Stebalien (commented):
Let's resolve these final comments in our hack time and get this merged. Then we need to write some tests. We can also try to simplify/streamline even more, but let's get this merged first.

src/Payments.sol
aarshkshah1992: @Stebalien I agree. I am simplifying it in the current PR itself.

src/Payments.sol
```diff
@@ -136,9 +160,68 @@ contract Payments is
...
+        // Settle account lockup to ensure we're up to date
+        settleAccountLockup(payer);
+
+        // Verify that the account is fully settled up to the current epoch
+        // This ensures that the client has enough funds locked to settle the rail upto
+        // and including (termination epoch aka current epoch + rail lockup period)
+        require(
+            payer.lockupLastSettledAt >= block.number,
+            "cannot terminate rail: failed to settle account lockup completely"
+        );
```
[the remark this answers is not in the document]
aarshkshah1992: Done.

src/Payments.sol
```diff
@@ -448,29 +709,75 @@ contract Payments is
...
         if (startEpoch >= maxSettlementEpoch) {
-            return (0, startEpoch, "already settled upto requested epoch");
+            return (
+                0,
+                startEpoch,
+                string(
+                    abi.encodePacked(
+                        "already settled up to maxSettlementEpoch epoch ",
+                        Strings.toString(maxSettlementEpoch)
+                    )
```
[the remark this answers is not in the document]

src/Payments.sol
```diff
@@ -448,29 +709,75 @@ contract Payments is
...
-        // nothing to settle (already settled or zero-duration)
+        // Nothing to settle (already settled or zero-duration)
         if (startEpoch >= maxSettlementEpoch) {
-            return (0, startEpoch, "already settled upto requested epoch");
+            return (
+                0,
+                startEpoch,
+                string(
+                    abi.encodePacked(
+                        "already settled up to maxSettlementEpoch epoch ",
```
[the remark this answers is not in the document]

src/Payments.sol
```diff
@@ -667,50 +1047,92 @@ contract Payments is
...
-                : 0;
+            return currentEpoch;
+        }
 
-            if (availableFunds == 0) {
-                return (false, account.lockupLastSettledAt);
-            }
+        require(
+            account.funds >= account.lockupCurrent,
+            "failed to settle: insufficient funds to cover lockup"
```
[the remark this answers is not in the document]

src/Payments.sol
```diff
@@ -233,49 +316,106 @@ contract Payments is
...
+
+    function modifyTerminatedRailLockup(
+        uint256 railId,
+        uint256 period,
+        uint256 lockupFixed
+    ) internal {
         Rail storage rail = rails[railId];
+
+        require(
+            period == rail.lockupPeriod && lockupFixed < rail.lockupFixed,
```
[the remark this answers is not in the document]

src/Payments.sol
```diff
@@ -393,45 +571,128 @@ contract Payments is
...
-    function validateAndModifyRateChangeApproval(
+    function calculateEffectiveLockupPeriod(
+        uint256 railId,
         Rail storage rail,
-        OperatorApproval storage approval,
+        Account storage payer
+    ) internal view returns (uint256 effectiveLockupPeriod) {
+        if (!isRailTerminated(railId)) {
+            return
+                rail.lockupPeriod - (block.number - payer.lockupLastSettledAt);
```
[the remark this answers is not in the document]
aarshkshah1992: This can only happen if a rail is in debt because we settle account lockup at the beginning of a rate change. 

Fixing this by setting effective lockup to 0 if rail is in debt because we disallow rate changes anyways for in-debted rails.

src/Payments.sol
aarshkshah1992: @Stebalien Have simplified this and it looks much cleaner now. Please take a look.

src/Payments.sol
```diff
@@ -296,86 +434,122 @@ contract Payments is
...
+        uint256 newRate
+    ) internal {
+        // If rate hasn't changed, nothing to do
+        if (newRate == oldRate) {
+            return;
+        }
+
+        // If there is no arbiter, settle the rail immediately
+        if (rail.arbiter == address(0)) {
+            (, uint256 settledUpto, ) = settleRail(railId, block.number, false);
```
Stebalien, line 534: change this to take a rail?

Stebalien: (trying to avoid passing rail + rail id everywhere)

Stebalien (approved):

src/Payments.sol
aarshkshah1992: This is the only place we pass a `railId` now because `settleRail` is a public function that can be called by EOAs as well. And they only need to pass the `railId`.

src/Payments.sol
```diff
@@ -217,6 +220,14 @@ contract Payments is
...
+
+        // Update operator approval rate usage
+        require(
+            approval.rateUsage >= rail.paymentRate,
+            "invariant violation: operator rate usage must be at least the rail payment rate"
+        );
+        approval.rateUsage -= rail.paymentRate;
+
         require(
             payer.lockupCurrent <= payer.funds,
```
Stebalien, line 232: is this the right check? We're not touching `lockupCurrent`.

Stebalien (approved):

src/Payments.sol
Stebalien: Ah, I see.

src/Payments.sol
aarshkshah1992: Done. You are right.

Stebalien (approved):

PR #10, opened 2025-03-12
remove approveOperator in favor of setOperatorApproval

Stebalien:
These are basically the same function, but the latter makes it possible to unapprove the operator. We might want to think about naming here, but I think we can delete the more-specific method.

### DIFF
--- a/src/Payments.sol
+++ b/src/Payments.sol
@@ -143,23 +143,6 @@ contract Payments is
         _;
     }
 
-    function approveOperator(
-        address token,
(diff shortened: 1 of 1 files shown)


aarshkshah1992:
lgtm.  We can  probably remain `setOperatorApproval` to `createOrUpdateOperatorApproval` or something similar to emphasise that it can create a new operator approval as well.

aarshkshah1992 (approved):

PR #12, opened 2025-03-12
don't call a nonReentrant function from a nonReentrant function

Stebalien:


Stebalien:
I _think_ we can just remove the `nonReentrant` restriction here. I assume the call to `safeTransfer` causes the entire function to revert on failure?

aarshkshah1992:
@Stebalien Not sure what this PR gets us. We can call non-reentrant functions in the same contract from another non-reentrant function.

aarshkshah1992 (approved):

Stebalien:
It likely won't hurt, but I don't want users to be surprised. We can reconsider it later.

Issue #4, opened 2025-03-04
When modifying rail payment, don't queue rate change when settled


Stebalien:
Specifically, handle the case where the user:

1. Settles the rail.
2. Changes the rate.

In this case, we shouldn't put anything into the queue. I.e., when `rail.settledUpTo` is the current epoch.

aarshkshah1992:
Closed by https://github.com/FilOzone/payments/pull/19.

PR #19, opened 2025-03-13
feat: rate change fixes

### LINKED ISSUE #16: Implicitly switch to a "0" rate one lockup period after termination
Right now, calls to `modifyRailPayment` after the lockup period expires revert if we've passed the "last settlement epoch". To simplify the mental model of how all this works, I think we should change this such that the rail's rate is implicitly set to 0 at this point. From the user's perspective, this means:

1. Asking for the rate on a terminated rail past this last settlement epoch will return 0.
2. Calls to `modifyRailPayment` will _succeed_ as long as the new rate is 0 and the one-time payment is non-existent. Yes, this is stupid. But it makes the docs simpler.

aarshkshah1992:
closes #16 and #4 .

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,5 @@ docs/
 
 # Dotenv file
 .env
+
+.DS_Store
--- a/src/Payments.sol
+++ b/src/Payments.sol
@@ -525,6 +525,12 @@ contract Payments is
             return;
         }
 
+        // No need to settle the rail or enqueue the rate change if the rail has already been settled upto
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -12,3 +12,5 @@ docs/
 
 # Dotenv file
 .env
+
+.DS_Store
```
Stebalien, line 16: Not an issue to have this here, but I recommend that you create a global gitignore file.

PR #21, opened 2025-03-13
Allow lockup reduction for non-terminated rails not in debt even if account lockup is not fully settled

### LINKED ISSUE #14: Allow reducing the lockup period when the account isn't fully funded but the rail isn't in debt.

aarshkshah1992:
Closes #14 .

@Stebalien This PR allows operators to reduce **_BOTH fixed lockup and lockup period_** if Rail is NOT in debt.

### DIFF
--- a/src/Payments.sol
+++ b/src/Payments.sol
@@ -390,19 +390,47 @@ contract Payments is
             rail.from
         ][rail.operator];
 
-        // Ensure account lockup is fully settled up to the current epoch
-        uint256 lockupSettledUpto = settleAccountLockup(payer);
         require(
-            lockupSettledUpto == block.number,
-            "cannot modify active rail lockup: client funds insufficient for current account lockup settlement"
+            !isRailInDebt(rail, payer),
+            "cannot modify rail lockup: rail is in debt"
         );
 
-        // Calculate current (old) lockup - for active rails this is straightforward
+        // Settle account lockup as much as possible
+        uint256 lockupSettledUpto = settleAccountLockup(payer);
+
+        // Only require full settlement if increasing period or fixed lockup
+        if (period > rail.lockupPeriod || lockupFixed > rail.lockupFixed) {
+            require(
+                lockupSettledUpto == block.number,
+                "cannot increase lockup: client funds insufficient for current account lockup settlement"
+            );
+        } else {
+            // When reducing period, ensure we still cover all unsettled epochs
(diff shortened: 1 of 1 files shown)


src/Payments.sol
```diff
@@ -390,19 +390,47 @@ contract Payments is
...
         ][rail.operator];
 
-        // Ensure account lockup is fully settled up to the current epoch
-        uint256 lockupSettledUpto = settleAccountLockup(payer);
         require(
-            lockupSettledUpto == block.number,
-            "cannot modify active rail lockup: client funds insufficient for current account lockup settlement"
+            !isRailInDebt(rail, payer),
+            "cannot modify rail lockup: rail is in debt"
         );
```
Stebalien, line 396: Remove this? We're checking this below and we can't check it until we've settled lockup.

src/Payments.sol
```diff
@@ -390,19 +390,47 @@ contract Payments is
...
+
+        // Only require full settlement if increasing period or fixed lockup
+        if (period > rail.lockupPeriod || lockupFixed > rail.lockupFixed) {
+            require(
+                lockupSettledUpto == block.number,
+                "cannot increase lockup: client funds insufficient for current account lockup settlement"
+            );
+        } else {
+            // When reducing period, ensure we still cover all unsettled epochs
+            if (period < rail.lockupPeriod) {
```
Stebalien, line 409: nit: `else if`

src/Payments.sol
```diff
@@ -390,19 +390,47 @@ contract Payments is
...
+        }
+
+        // Calculate effective lockup period for the old period
+        uint256 oldEffectiveLockupPeriod = rail.lockupPeriod -
+            (block.number - payer.lockupLastSettledAt);
+
+        // Calculate effective lockup period for the updated period
+        uint256 newEffectiveLockupPeriod = period -
+            (block.number - payer.lockupLastSettledAt);
+
```
Stebalien, line 424: What happens if we reduce the fixed lockup but the rail is way in debt? I think these statements will revert.

aarshkshah1992:
@Stebalien Have addressed your review.

src/Payments.sol
aarshkshah1992: Done.

src/Payments.sol

aarshkshah1992:
Closing in favour of https://github.com/FilOzone/payments/pull/25.

PR #24, opened 2025-03-14
Simplify one time payments 

### LINKED ISSUE #22: Standardizing how we settle lockup
Ideally we'd turn it into modifier (possibly settling at both the top and the bottom of every function. But, we may also want to:

1. Enqueue a rate-change to 0 for terminated rails past the last settlement epoch (so we don't have to reason about this case).
2. Assert that the rail is active/valid.
3. Settle lockup as a post-condition as well.
4. Possibly assert the lockup state (pre/post)?
5. I'm wondering if it's even possible to track operator lockup allowances here. I.e., it may be possible to just (a) record the before state (b) let the operator do whatever (c) check if the changes are valid given the operator's allowance. But that's a stretch.

I think we might also want to rename it (easily confused with "settle rail"). Maybe just call it `updateRail`?

### LINKED ISSUE #23: Simplify accounting of one-time payments
This code:

https://github.com/FilOzone/payments/blob/a557ca3dd2a4e0f86b3312aea783fd28a91a37ad/src/Payments.sol#L664-L674

I think it's kind-of correct now, but it's still a bit confusing (and not 100% correct). From what I can tell, we're:

1. Removing the one-time payment from the operator's approval tracking in the call to `updateOperatorLockupTracking`.
2. Adding it back in the second part.

That will ensure that:

1. We remove the one-time payment from the per-rail fixed lockup.
2. We _don't_ let the operator re-spend said payment.

On the other hand, it's a bit awkward because the operator's fixed lockup will have to keep going up and up to cover new one-time payments. Ideally, that second step would actually subtract the one-time payment from the operator's _allowance_, I think? But that's where this gets real complicated. I'm not 100% sure.

aarshkshah1992:
Closes #23.

### DIFF
--- a/src/Payments.sol
+++ b/src/Payments.sol
@@ -70,6 +70,7 @@ contract Payments is
         uint256 lockupAllowance;
         uint256 rateUsage; // Track actual usage for rate
         uint256 lockupUsage; // Track actual usage for lockup
+        uint256 oneTimePaymentsUsed; // Track one-time payments made by this operator
     }
 
     // Counter for generating unique rail IDs
@@ -360,9 +361,12 @@ contract Payments is
         // 1. Period remains unchanged (enforced by the require statement above)
         // 2. The rate-based portion of the lockup doesn't change
         // 3. The only thing changing is the fixed lockup, which is being reduced
-
-        // Update operator allowance - reduce usage by the exact reduction amount
-        approval.lockupUsage -= lockupReduction;
+        updateOperatorLockupTracking(
+            approval,
+            rail.lockupFixed,
+            lockupFixed,
+            0
+        );
 
         // Update payer's lockup - subtract the exact reduction amount
         require(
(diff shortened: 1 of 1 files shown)


src/Payments.sol
```diff
@@ -70,6 +70,7 @@ contract Payments is
         uint256 lockupAllowance;
         uint256 rateUsage; // Track actual usage for rate
         uint256 lockupUsage; // Track actual usage for lockup
+        uint256 oneTimePaymentsUsed; // Track one-time payments made by this operator
```
Stebalien, line 73: Any reason not to reduce the allowance instead?

Stebalien: I.e., reduce the usage _and_ allowance at the same time (clipped at 0)?

aarshkshah1992: Yeah that's a better idea. Closing this.

aarshkshah1992:
Closed in favour of https://github.com/FilOzone/payments/pull/25.

PR #25, opened 2025-03-17
feat: move to an invariants model to verify correct state mutation

### LINKED ISSUE #14: Allow reducing the lockup period when the account isn't fully funded but the rail isn't in debt.

### LINKED ISSUE #15: Consider enqueuing rate-changes for rails without arbiters
It makes it easier to explain/understand if changing the rate never settles.

### LINKED ISSUE #22: Standardizing how we settle lockup

### LINKED ISSUE #23: Simplify accounting of one-time payments

aarshkshah1992:
Closes #22.
Closes #23.
Closes #15.
Closes #14.

### DIFF
--- a/src/Payments.sol
+++ b/src/Payments.sol
@@ -100,7 +100,6 @@ contract Payments is
     ) internal override onlyOwner {}
 
     modifier validateRailActive(uint256 railId) {
-        require(rails[railId].from != address(0), "rail does not exist");
         require(rails[railId].isActive, "rail is inactive");
         _;
     }
@@ -439,7 +438,7 @@ contract Payments is
         // Only require full settlement if increasing period or fixed lockup
         if (period > rail.lockupPeriod || lockupFixed > rail.lockupFixed) {
             require(
-                payer.lockupSettledUpto == block.number,
+                payer.lockupLastSettledAt == block.number,
                 "cannot increase lockup: client funds insufficient for current account lockup settlement"
             );
         } else if (period < rail.lockupPeriod) {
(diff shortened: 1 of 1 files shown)


src/Payments.sol
```diff
@@ -1091,7 +1115,7 @@ contract Payments is
 
     // attempts to settle account lockup up to and including the current epoch
     // returns the actual epoch upto and including which the lockup was settled
-    function settleAccountLockup(
+    function _settleAccountLockup(
```
Stebalien, line 1118: In general, please leave renames to separate PRs.

Stebalien (commented):
Ok, so, I tried but there's simply way too much going on in this PR for me to give it a good review. Let's break this into separate feature PRs then top it off with the refactor. Or you can go about it the other way around, whatever is easier.

PR #29, opened 2025-03-19
invariants PR #1: allow operator to reduce lockup period and fixed lockup if rail is not in debt even if account is not fully settled and use only the nonreentrant modifier for reentrancy protection

### LINKED ISSUE #14: Allow reducing the lockup period when the account isn't fully funded but the rail isn't in debt.

### LINKED ISSUE #26: Do we still need `Rail.isActive`?
What if we:

1. Check if it has a sender, etc.
2. Zero the entire thing when we're done (frees space).

### LINKED ISSUE #27: Per-rail reentrancy guard isn't quite correct
We have a per-rail reentrancy guard (`noRailModificationInProgress`) and a global one (`nonReentrant`). Unfortunately, they don't actually work together which means:

1. We can enter calls marked `noRailModificationInProgress` when `nonReentrant` is active.
2. We can enter calls marked `nonReentrant` when `noRailModificationInProgress` is active for a given rail.

We either need to unify these somehow, or switch entirely to one or the other.

aarshkshah1992:
Closes #14
Closes #27 
Closes #26

### DIFF
--- a/src/Payments.sol
+++ b/src/Payments.sol
@@ -98,7 +98,10 @@ contract Payments is
     ) internal override onlyOwner {}
 
     modifier validateRailActive(uint256 railId) {
-        require(rails[railId].from != address(0), "rail does not exist or is inactive");
+        require(
+            rails[railId].from != address(0),
+            "rail does not exist or is inactive"
+        );
         _;
     }
 
@@ -1188,11 +1191,7 @@ contract Payments is
     ) internal view returns (bool) {
         return block.number > payer.lockupLastSettledAt + rail.lockupPeriod;
     }
-    
-    /**
-     * @dev Internal function to zero out all fields of a rail to mark it as inactive.
-     * Setting rail.from to address(0) is the main indicator that a rail is inactive.
-     */
+
     function _zeroOutRail(Rail storage rail) internal {
         rail.token = address(0);
         rail.from = address(0); // This now marks the rail as inactive
@@ -1204,7 +1203,7 @@ contract Payments is
         rail.lockupPeriod = 0;
         rail.settledUpTo = 0;
         rail.terminationEpoch = 0;
-        
+
         // Clear the rate change queue
         rail.rateChangeQueue.clear();
     }
--- a/src/Payments.sol
+++ b/src/Payments.sol
@@ -48,7 +48,6 @@ contract Payments is
     }
 
     struct Rail {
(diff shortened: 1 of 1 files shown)


src/Payments.sol
```diff
@@ -1175,6 +1191,22 @@ contract Payments is
...
+        rail.token = address(0);
+        rail.from = address(0); // This now marks the rail as inactive
+        rail.to = address(0);
+        rail.operator = address(0);
+        rail.arbiter = address(0);
+        rail.paymentRate = 0;
+        rail.lockupFixed = 0;
+        rail.lockupPeriod = 0;
+        rail.settledUpTo = 0;
+        rail.terminationEpoch = 0;
```
Stebalien, line 1205: I think you can do this as follows:

```solidity
Rail storage emptyRail;
rail = emptyRail
```

But I'm not 100% sure? You'll probably need to clear the queue at least but... I _think_ the queue should be empty at this point so I'd just assert that instead.

src/Payments.sol
```diff
@@ -378,19 +370,46 @@ contract Payments is
...
+            require(
+                lockupSettledUpto == block.number,
+                "cannot increase lockup: client funds insufficient for current account lockup settlement"
+            );
+        } else if (period < rail.lockupPeriod) {
+            // When reducing period, ensure we still cover all unsettled epochs
+            require(
+                payer.lockupLastSettledAt + period >= block.number,
+                "cannot reduce lockup period below what's needed for unsettled epochs"
+            );
```
Stebalien, line 393: 1. I think this check is redundant with `isRailInDebt`. I'd actually just keep this one and remove the `isRailInDebt` as, IMO, this is clearer. But up to you.
2. I don't think either check is actually necessary, but that's not what I wrote in the initial issue. If/when the rail is in debt, the effective lockup period is 0 so we won't unlock/relock any funds, but we can still reduce the period. I'm happy to leave this as-is for now.

Stebalien (approved):

src/Payments.sol
Stebalien: Issue filed as https://github.com/FilOzone/payments/issues/34.

src/Payments.sol
Stebalien: https://github.com/FilOzone/payments/issues/36

PR #30, opened 2025-03-19
invariants PR #2: lay groundwork for invariant checks and refactor modifyRailRate

aarshkshah1992:
Have raised this on top of https://github.com/FilOzone/payments/pull/29. Please review that first.

### DIFF
--- a/src/Payments.sol
+++ b/src/Payments.sol
@@ -456,28 +456,49 @@ contract Payments is
         Account storage payee = accounts[rail.token][rail.to];
 
         uint256 oldRate = rail.paymentRate;
+        bool isTerminated = isRailTerminated(rail);
 
         // Settle the payer's lockup to account for elapsed time
         uint256 lockupSettledUpto = settleAccountLockup(payer);
 
         // Validate rate changes based on rail state and account lockup
-        validateRateChangeRequirements(
-            rail,
-            payer,
-            oldRate,
-            newRate,
-            lockupSettledUpto,
-            oneTimePayment
-        );
+        if (isTerminated) {
+            if (block.number >= maxSettlementEpochForTerminatedRail(rail)) {
+                return
+                    modifyPaymentForTerminatedRailBeyondLastEpoch(
+                        rail,
+                        newRate,
+                        oneTimePayment
+                    );
+            }
+
+            require(
+                newRate <= oldRate,
+                "failed to modify rail: cannot increase rate on terminated rail"
+            );
+        } else {
+            validateRateChangeRequirementsForNonTerminatedRail(
+                rail,
+                payer,
+                oldRate,
+                newRate,
+                lockupSettledUpto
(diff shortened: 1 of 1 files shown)


src/Payments.sol
```diff
@@ -1207,6 +1210,22 @@ contract Payments is
         // Clear the rate change queue
         rail.rateChangeQueue.clear();
     }
+
+    function withdrawToInternal(
+        address token,
+        address to,
+        uint256 amount
+    ) internal {
+        Account storage account = accounts[token][msg.sender];
```
Stebalien, line 1219: this needs to settle account lockup (for now)

Stebalien:
This should really be multiple independent PRs:

1. One to lift a bunch of requirements into modifiers.
2. One per significant change after that, with descriptions of what's changing.

As-is, this is still hard to review. Not because it's a big PR, but because it's hard to tell what should be happening.

When I review a PR, I look at:

1. The description of what should be changing.
2. The code before.
3. The code after.

And make sure they all match up.

In this case, I have absolutely no idea what I should be looking for which means I'm looking at the before/after and trying to figure out why these changes might have been made.

One thing that can also help is to leave code review comments on your own PR explaining why you made some interesting changes. If the PR is small, this is usually covered in the PR description. But this can help if you need to make other small changes that are tangentially related to the main change.

Stebalien:
Note: initial PRs are different because I'm reviewing _everything_ all at once as if it were fresh code because it usually is (or, at least, is mostly new code). The tricky part is when you're modifying existing code. I can't just look at the entire smart contract as if it's new because it isn't and I'm already familiar with what it should do, so I'll make assumptions and get things wrong.

src/Payments.sol
```diff
@@ -447,28 +456,49 @@ contract Payments is
...
         // Validate rate changes based on rail state and account lockup
-        validateRateChangeRequirements(
-            rail,
-            payer,
-            oldRate,
-            newRate,
-            lockupSettledUpto,
-            oneTimePayment
-        );
+        if (isTerminated) {
```
Stebalien, line 465: I really want to see this change in a separate PR where I only have to look at changes related to this and I know what exactly we're trying to achieve. Otherwise, it's just moving code around and I have no idea what's going on.

Stebalien:
Note: this feedback applies to the previous PR as well, but there I was able to mostly tease apart the different changes. It's fine to put things into larger PRs but, when you do that, changes at least need to be in separate commits with a clear progression; not the progression in how you wrote the code, but laid out as a story to the reviewer.

aarshkshah1992:
@Stebalien This is all good feedback. Thanks a lot. Need to be more mindful of our async workflow here. Breaking it down and will be careful with PRs going forward.

aarshkshah1992:
Closing in favour of the new PRs that break this down.

PR #35, opened 2025-03-19
simplify zeroing the payment rail

Stebalien:
1. Assert that the rate-change queue is empty instead of emptying it.
2. Zero the rail all at once.


Stebalien:
Ugh. We can't do this because the rail has storage fields.

PR #40, opened 2025-03-20
PR #2: refactor: simplify withdrawal function to remove redundant checks

aarshkshah1992:
This PR refactors both the withdrawal methods to simplify the checks they make.

### DIFF
--- a/src/Payments.sol
+++ b/src/Payments.sol
@@ -278,6 +278,13 @@ contract Payments is
     ) internal {
         Account storage account = accounts[token][msg.sender];
 
+        // Settle account lockup to ensure we're up to date before withdrawal
+        uint256 settledUntil = settleAccountLockup(account);
+        require(
+            settledUntil == block.number,
+            "withdrawal requires account lockup to be fully settled"
+        );
+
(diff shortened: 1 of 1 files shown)


src/Payments.sol
```diff
@@ -276,23 +276,14 @@ contract Payments is
         address to,
         uint256 amount
     ) internal {
-        require(token != address(0), "token address cannot be zero");
-        require(to != address(0), "recipient address cannot be zero");
-
-        Account storage acct = accounts[token][msg.sender];
-
-        uint256 settleEpoch = settleAccountLockup(acct);
```
Stebalien, line 284: I think we still need to settle here.

Stebalien (changes requested):

src/Payments.sol
aarshkshah1992: We take care of that using a modifier in https://github.com/FilOzone/payments/pull/33.

But have reverted this change so we can merge this in main.

Stebalien: I know, but each PR needs to stand on it's own. Otherwise I can't review them.

Stebalien (approved):

PR #42, opened 2025-03-20
feat: assert that rate change queue is empty post final settlement before zeroing out the rail

### LINKED ISSUE #36: Assert that there are no rate-changes when zeroing the rail
Right now we check, but we should consider asserting. At that point, we shouldn't have any unsettled rate changes.

aarshkshah1992:
Closes #36.

### DIFF
--- a/src/Payments.sol
+++ b/src/Payments.sol
@@ -1198,6 +1198,8 @@ contract Payments is
             rail.rateChangeQueue.isEmpty(),
             "rate change queue must be empty post full settlement"
         );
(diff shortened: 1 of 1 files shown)


Stebalien (approved):
Note: you don't need to clear this at the bottom now.

aarshkshah1992:
@Stebalien Yeah made the change.

Stebalien:
Ah, I meant we don't need to clear the queue. But I was wrong and you're right: we need to clear it to set the internal variables to zero.

PR #50, opened 2025-03-24
Add a public getRail method

Stebalien:


Stebalien:
I've broken this out into a separate PR to make the history a bit cleaner (not necessary at this point, but it's nice to be able to see when important features were added versus tests).

I've also added a restriction that the rail must exist as a minor sanity check.

PR #52, opened 2025-03-25
Fix bugs discovered during testing PR #2: Allow Rail settlement and foreclosure

aarshkshah1992:
@Stebalien 

- I don't think we should allow people to withdraw the rail lockup during normal settlement. It messes up all of our accounting. When you settle a rail in the normal cause of events, you can only withdraw upto and including the account lockup.

- If you want to withdraw the rail "future" lockup as well (might not be the "future" for a rail in debt) -> you should foreclose the rail and be done with it as you are essentially saying that you don't trust the client to deposit more funds anyways.

### DIFF
--- a/src/Payments.sol
+++ b/src/Payments.sol
@@ -827,7 +827,7 @@ contract Payments is
         // Handle terminated and fully settled rails that are still not finalised
         if (
             isRailTerminated(rail) &&
-            rail.settledUpTo >= maxSettlementEpochForTerminatedRail(rail)
+            rail.settledUpTo >= rail.endEpoch
         ) {
             finalizeTerminatedRail(rail, payer);
             return (0, rail.settledUpTo, "rail fully settled and finalized");
@@ -840,7 +840,7 @@ contract Payments is
         } else {
             maxSettlementEpoch = min(
                 untilEpoch,
-                maxSettlementEpochForTerminatedRail(rail)
+                rail.endEpoch
             );
         }
 
--- a/src/Payments.sol
+++ b/src/Payments.sol
@@ -815,7 +815,7 @@ contract Payments is
             "terminated rail can only be settled without arbitration after max settlement epoch"
         );
 
-        return settleRailInternal(railId, maxSettleEpoch, true);
+        return settleRailInternal(railId, maxSettleEpoch, true, false);
     }
 
     /// @notice Settles payments for a rail up to the specified epoch. Settlement may fail to reach the target epoch if either the client lacks the funds to pay up to the current epoch or the arbiter refuses to settle the entire requested range.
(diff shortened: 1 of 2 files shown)


src/Payments.sol
```diff
@@ -838,13 +838,81 @@ contract Payments is
...
+    }
+
+    /// @notice Settles and forecloses a rail. Can only be called by the operator or payee.
+    /// This method settles the rail, credits available funds to payee (up to account.lockupSettledAt + rail.lockupPeriod),
+    /// then finalizes the rail and marks it as inactive.
+    /// @param railId The ID of the rail to settle and foreclose.
+    /// @return totalSettledAmount The total amount settled and transferred.
+    /// @return finalSettledEpoch The epoch up to which settlement was actually completed.
+    /// @return note Additional information about the settlement.
+    function settleAndForecloseRail(
```
aarshkshah1992, line 851: only the operator should be able to call this

src/Payments.sol
```diff
@@ -838,13 +838,81 @@ contract Payments is
...
+        uint256 railId
+    )
+        external
```
aarshkshah1992, line 854: this should work only upto and including the current block and only if account lockup is NOT fully settled.

Stebalien:
I want to think about this API a bit:

1. I'd prefer to foreclose _then_ settle separately.
2. IMO, foreclosure is something the operator (not the payee) should handle as part of force-canceling deals.
3. We need to handle the case where the arbitrator won't fully settle the rail _yet_. Separating foreclosure & settling the foreclosed rail will help here.
4. I really think there is a way to combine foreclosure and termination. I'm trying to limit the number of mechanisms users have to consider.

I'd also like to spend some time playing around with the failing test.

aarshkshah1992:
@Stebalien 

- Removed the `isRailInDebt() helper` given our new design of only allowing settling a non-terminated rail upto the account lockup settlement epoch. If account lockup is NOT fully settled, ALL of it's rails are now in debt.
- Combine termination and foreclosure into one operation i.e. "rail termination":
   - Only the operator can terminate a rail if account lockup is NOT fully settled
   - Client can ONLY terminate a rail if account lockup is fully settled
- Settling a non-terminated rail only settles upto and including the account lockup settlement epoch (<= block number)   
- Settling a terminated rail settles upto and including min (user given input epoch, block number, rail.endEpoch i.e. rail termination epoch + rail lockup period).

src/Payments.sol
```diff
@@ -608,15 +598,13 @@ contract Payments is
             }
 
             require(
-                newRate <= oldRate,
-                "failed to modify rail: cannot increase rate on terminated rail"
+                newRate == oldRate,
```
aarshkshah1992, line 601: this really simplifies reasoning about correctness without losing much in UX and I'd to keep it this way.

src/Payments.sol
```diff
@@ -608,15 +598,13 @@ contract Payments is
...
+                "failed to modify rail: cannot change rate on terminated rail"
             );
         } else {
-            validateRateChangeRequirementsForNonTerminatedRail(
-                rail,
-                payer,
-                oldRate,
-                newRate
+            require(
+                isAccountLockupFullySettled(payer) || newRate == oldRate,
```
aarshkshah1992, line 606: same as above.

Stebalien: This I'm fine with (for now). I _want_ to be able to reduce the lockup in this case, but I think we can live with this.

src/Payments.sol
Stebalien: We _need_ to be able to reduce the rate after termination, or we need to be able to reduce the rate before termination but while in-debt and/or behind lockup.

Basically, we need to support the following:

1. User falls behind on lockup.
2. SP asks the operator to cancel all deals.
3. The operator must be able to stop funds on the rail _at this point_.

That means either (a) reducing the rate to zero before terminating or (b) need to be able to terminate and then reduce the rate to zero .

src/Payments.sol
```diff
@@ -608,15 +598,13 @@ contract Payments is
             }
```
Stebalien, line 598: Docs will need to be updated in any case.

src/Payments.sol
Stebalien: We _can_ do this with arbitration, but that's really terrible UX. Basically, I think the model needs to be:

1. If an account is behind on lockup, all active rails cannot be touched except that rails can be terminated and one-time payments can be made (really, we don't need the latter).
2. Once a rail has been terminated, the rail is "unlocked" and can be modified again (with restrictions).

For terminated rails, we use completely separate logic:

1. We can set the lockup period to 0 and simply use the end epoch as our source of truth. In theory, we could allow altering the end epoch but that's a future problem.
2. We remove the rail from the account's lockup rate, no more funds will be locked to the rail.
3. The lockup period cannot be changed.
4. If the rail's rate is reduced, we unlock (`rateDelta * (endEpoch - currentEpoch)`).

src/Payments.sol
```diff
@@ -861,32 +824,25 @@ contract Payments is
...
+            return (0, rail.settledUpTo, "rail fully settled and finalized");
         }
 
-        uint256 maxLockupSettlementEpoch = payer.lockupLastSettledAt +
-            rail.lockupPeriod;
-        uint256 maxSettlementEpoch = min(untilEpoch, maxLockupSettlementEpoch);
+        // Calculate the maximum settlement epoch based on account lockup
+        uint256 maxSettlementEpoch;
+        if (!isRailTerminated(rail)) {
+            maxSettlementEpoch = min(untilEpoch, payer.lockupLastSettledAt);
```
Stebalien, line 839: I need to stop working on FVM tech debt and take some time tomorrow to try to actually fix the underlying issue here. The fact that we can't settle these locked funds is really annoying me, there's no technical reason this isn't possible.

src/Payments.sol
```diff
@@ -861,32 +824,25 @@ contract Payments is
...
-                    "rail fully settled and finalized"
-                );
-            }
-
-            // For terminated but not fully settled rails, limit settlement window
-            untilEpoch = min(untilEpoch, maxTerminatedRailSettlementEpoch);
+        // Handle terminated and fully settled rails that are still not finalised
+        if (
+            isRailTerminated(rail) &&
+            rail.settledUpTo >= maxSettlementEpochForTerminatedRail(rail)
```
Stebalien, line 830:
```suggestion
            rail.settledUpTo >= rail.endEpoch
```

(calling `maxSettlementEpochForTerminatedRail` is kind of overkill here)

src/Payments.sol
Stebalien: (and maybe I'm wrong, but this is a case where I need to prove this to myself)

src/Payments.sol
```diff
@@ -861,32 +824,25 @@ contract Payments is
...
-            rail.lockupPeriod;
-        uint256 maxSettlementEpoch = min(untilEpoch, maxLockupSettlementEpoch);
+        // Calculate the maximum settlement epoch based on account lockup
+        uint256 maxSettlementEpoch;
+        if (!isRailTerminated(rail)) {
+            maxSettlementEpoch = min(untilEpoch, payer.lockupLastSettledAt);
+        } else {
+            maxSettlementEpoch = min(
+                untilEpoch,
+                maxSettlementEpochForTerminatedRail(rail)
```
Stebalien, line 843:
```suggestion
                rail.endEpoch
```

src/Payments.sol
aarshkshah1992: Sure. I have thought a lot about this, have written tests to understand this better and doing this really simplifies accounting without sacrificing much in the way of UX. 

I am sure we can come up with **_more_** arithmetic to be able to do this but it adds one more dimension of complexity to an already complex contract and I would like to first give this contract to people and complain about not being able to get locked funds without rail termination before we introduce this complexity.

aarshkshah1992:
@Stebalien Made the change. Can you review again today ?

Stebalien (approved):
I think this is right (but it's getting to the point where I'll need to do a fresh audit top-to-bottom).

PR #58, opened 2025-04-11
Flesh out the README and improve some of the docs/error messages

Stebalien:
This PR mostly fleshes out the README, but I've also modified some of the error messages and internal docs to use the phrase "fully funded" instead of walking about "settling lockup" because I think the former will be easier for the user to understand.


Stebalien:
Docs drawn from:

1. The code & method documentation.
2. The example flow in the payment design iteration notion doc.

With some help from claude.

aarshkshah1992 (approved):

PR #59, opened 2025-04-14
[WIP] Payment changes based on PDP integration

aarshkshah1992:
WIP PR for making changes to the payments contract based on PDP integration.

### DIFF
--- a/src/Payments.sol
+++ b/src/Payments.sol
@@ -72,7 +72,7 @@ contract Payments is
     }
 
     // Counter for generating unique rail IDs
-    uint256 private _nextRailId;
+    uint256 private _nextRailId = 1;
 
     // token => owner => Account
     mapping(address => mapping(address => Account)) public accounts;
@@ -257,7 +257,6 @@ contract Payments is
                 lockupFixed: rail.lockupFixed,
                 settledUpTo: rail.settledUpTo,
                 endEpoch: rail.endEpoch
-
             });
     }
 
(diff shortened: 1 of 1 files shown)


src/Payments.sol
```diff
@@ -762,7 +761,7 @@ contract Payments is
         nonReentrant
         validateRailActive(railId)
         validateRailTerminated(railId)
-        onlyRailClient(railId)
+        onlyRailOperator(railId)
```
Stebalien, line 764: The point of this was to allow the client to bypass a broken operator/arbitrator (which will generally be the same entity). If the operator needs the ability to do this, I'd just let them change the arbiter.

src/Payments.sol
```diff
@@ -795,6 +794,7 @@ contract Payments is
         public
         nonReentrant
         validateRailActive(railId)
+        onlyRailOperator(railId)
```
Stebalien, line 797: Why? I think we definitely want any party to be able to do this. We can restrict it to client + operator + SP, but we don't want to wait on the operator.

src/Payments.sol
```diff
@@ -72,7 +72,7 @@ contract Payments is
     }
 
     // Counter for generating unique rail IDs
-    uint256 private _nextRailId;
+    uint256 private _nextRailId = 1;
```
Stebalien, line 75: nice catch!

aarshkshah1992:
@Stebalien Raising a new one with better more documented changes.

PR #62, opened 2025-04-22
feat: add getters to get railIds for payer and payee

aarshkshah1992:
Needed for Dapps. They always know what token they're dealing with and once they have the railIds -> can use the existing rail getter to get more information


aarshkshah1992:
Closed in favour of https://github.com/FilOzone/fws-payments/pull/64

PR #64, opened 2025-04-22
feat: add getters to get railIds for payer and payee #62

aarshkshah1992:
Needed for Dapps. They always know what token they're dealing with and once they have the railIds -> can use the existing rail getter to get more information

### DIFF
--- a/src/Payments.sol
+++ b/src/Payments.sol
@@ -907,18 +907,17 @@ contract Payments is
         uint256 startEpoch = rail.settledUpTo;
         // Nothing to settle (already settled or zero-duration)
         if (startEpoch >= maxSettlementEpoch) {
-            return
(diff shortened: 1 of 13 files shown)


src/Payments.sol
```diff
@@ -1466,6 +1485,96 @@ contract Payments is
...
+        }
+
+        // Create correctly sized final result array
+        RailInfo[] memory result = new RailInfo[](resultCount);
+
+        // Only copy if we have results (avoid unnecessary operations)
+        if (resultCount > 0) {
+            for (uint256 i = 0; i < resultCount; i++) {
+                result[i] = tempResults[i];
+            }
```
Stebalien, line 1573: Oh solidity... :disappointed:

PR #65, opened 2025-04-22
feat: add versioning

### LINKED ISSUE #63: Add versioning to payment contract implementation
As we are using UUPS to enable upgradable smart contracts, it is essential to maintain a clean versioning strategy so there is a clear trace of changes over time & allow clients know which version they are interacting with & how.

Consider adding the following to the implementation contract 

```solidity 
string public constant VERSION = "0.0.0";  // or a semver tag

function getVersion() public pure returns (string memory) {
    return VERSION;
}

```

jennijuju:
resolves #63

### DIFF
--- a/src/Payments.sol
+++ b/src/Payments.sol
@@ -39,6 +39,12 @@ contract Payments is
 {
     using SafeERC20 for IERC20;
     using RateChangeQueue for RateChangeQueue.Queue;
+    
+    string public constant VERSION = "0.0.0";
+    
+    function getVersion() public pure returns (string memory) {
+        return VERSION;
+    }
 
     struct Account {
         uint256 funds;


src/Payments.sol
```diff
@@ -39,6 +39,19 @@ contract Payments is
...
+    
+    event ContractUpgraded(string version, address implementation);
+    
+    function getVersion() public pure returns (string memory) {
+        return VERSION;
+    }
+    
+    function postUpgrade() external {
+        require(msg.sender == address(this), "Only callable via delegatecall from proxy");
+        emit ContractUpgraded(VERSION, address(this));
```
ZenGround0, line 53: Looking into this a big more this won't be the expected implementation value as written:   https://ethereum.stackexchange.com/questions/90201/what-is-the-value-of-addressthis-when-delegate-calling

address(this) is the proxy not the implementation.

ZenGround0: We'll want to grab this from the IMPLEMENTATIN_SLOT

ZenGround0: This one can be copied over here: https://github.com/FilOzone/pdp/pull/160/files

ZenGround0:
Assuming we end up with an upgradeable contract we'll just copy over what we did in PDP ( https://github.com/FilOzone/pdp/pull/160/files)

But for now let's hold off on this until we get a good handle on our upgradabiliy strategy.

rjan90:
Closing this PR, as we have currently have suggested that adopting a versioning scheme similar to UniSwap (e.g., FilecoinPayV1, V2, etc.) in the contract name itself, once we believe the contract has been stabalized, and we don’t anticipate further hard upgrades. 

See issue: https://github.com/FilOzone/filecoin-pay/issues/235

PR #66, opened 2025-04-22
feat: take both payment fee and operator commission for one time payments as well

aarshkshah1992:
Operator commission and payment fee should apply to one time payments as well.

### DIFF
--- a/src/Payments.sol
+++ b/src/Payments.sol
@@ -907,18 +907,17 @@ contract Payments is
         uint256 startEpoch = rail.settledUpTo;
         // Nothing to settle (already settled or zero-duration)
         if (startEpoch >= maxSettlementEpoch) {
-            return
-                checkAndFinalizeTerminatedRail(
-                    rail,
-                    payer,
-                    0,
-                    0,
-                    0,
-                    0,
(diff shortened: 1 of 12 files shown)


src/Payments.sol
```diff
@@ -805,18 +805,73 @@ contract Payments is
...
+            Account storage operatorAccount = accounts[token][operator];
+            operatorAccount.funds += operatorCommission;
+        }
+        
+        // Track platform fee
+        if (paymentFee > 0) {
+            // Check if this is the first fee for this token
+            if (accumulatedFees[token] == 0) {
+                feeTokens.push(token);
+            }
```
Stebalien, line 843: Ditto the other PR, this will break if we withdraw the funds down to zero.

Stebalien (approved):
Approved modulo the "draw down to zero" bug.

src/Payments.sol
aarshkshah1992: Fixed this bug in the other PR. Will rebase.

aarshkshah1992:
Replaced by https://github.com/FilOzone/fws-payments/pull/71.

PR #73, opened 2025-05-01
feat: add support for native FIL

### LINKED ISSUE #72: Native FIL support
(not an immediate priority but I wanted to write this down)

Adding support for native FIL shouldn't actually be that hard. We can just use `0x0` as the token address and everything stays the same except deposit/withdraw (which will only need slight modifications).

The simplest approach is to reuse all existing methods but:

1. Have the user specify 0x0 as the token address.
2. Modify deposit/withdraw to support native tokens.

However, we might want a separate _special_ native deposit/withdraw methods, just to be safe.

Stebalien:
This adds support for native FIL tokens by special-casing 0x0 on both deposit and withdraw.

This is the simplest possible approach. We might want to make it nicer by introducing separate withdraw/deposit methods but... I'm not actually sure if that's necessary or even nicer.

NOTE: 0x0 is the address of the Ethereum system actor and no contract will ever be deployed there, so this is always safe.

fixes #72

TODO:

- [ ] write tests


Stebalien:
@aarshkshah1992 is this a reasonable approach?

Stebalien:
(draft because it needs tests, docs, and better error messages)

src/Payments.sol
```diff
@@ -461,7 +463,12 @@ contract Payments is
             "insufficient unlocked funds for withdrawal"
         );
         account.funds -= amount;
-        IERC20(token).safeTransfer(to, amount);
+        if (token == address(0)) {
```
aarshkshah1992, line 466: this is really nice

Stebalien:
I've added basic deposit/withdrawal tests. The rest of the logic is identical, so this should be enough.

PR #74, opened 2025-05-02
Optimize Safe Loop Increments Using unchecked

pali101:
## Description

Replaced standard `i++` with `unchecked { ++i }` in for loops where overflow is provably impossible. Applies to loops over feeTokens, rail IDs, and result arrays.

## Rationale

The Solidity compiler adds an overflow check to all arithmetic operations by default. However, in controlled loops where the index variable is bounded and overflow is impossible, these checks are redundant. Using unchecked eliminates this overhead and aligns with widely accepted optimization practices.

## Impact

- No change to behavior
- Aligns code with Solidity gas optimization best practices
- All tests continue to pass (79/79)

## Gas Reports
[current_gas_reports.txt](https://github.com/user-attachments/files/20008403/current_gas_reports.txt):  Gas usage after replacing `i++` with `unchecked { ++i }`.
[previous_gas_reports.txt](https://github.com/user-attachments/files/20008402/previous_gas_reports.txt): Gas usage before the optimization.

### Gas Report Summary

| Metric                                 | Before        | After         | Delta     |
|----------------------------------------|---------------|----------------|-----------|
| **Payments contract deployment cost**  | 4,760,515     | 4,763,061      | +2,546     |
| **Payments contract size (bytes)**     | 21,661        | 21,673         | +12        |
| **upgradeImplementation() test**       | 4,858,568     | 4,861,125      | +2,557     |
| **getRailsForPayeeAndToken()**         | 108,809       | 108,911        | +102       |
| **getRailsForPayerAndToken()**         | 98,992        | 99,085         | +93        |
| **getAllAccumulatedFees()**            | 1,198,711     | 1,198,711      | 0          |
| **Total tests passing**                |  79 / 79     |  79 / 79      | –          |

Although there's a slight gas increase, it can be due to compiler layout shifts, not inefficiencies. Using unchecked skips overflow checks in safe loops, boosts performance in complex iterations, and prepares the code for future optimizations.

### DIFF
--- a/src/Payments.sol
+++ b/src/Payments.sol
@@ -1575,10 +1575,13 @@ contract Payments is
         count = feeTokens.length;
         tokens = new address[](count);
         amounts = new uint256[](count);
-        for (uint256 i = 0; i < count; i++) {
+        for (uint256 i = 0; i < count; ) {
             address token = feeTokens[i];
             tokens[i] = token;
             amounts[i] = accumulatedFees[token];
+            unchecked {
+                ++i;
+            }
         }
 
         return (tokens, amounts, count);
@@ -1631,12 +1634,17 @@ contract Payments is
         RailInfo[] memory tempResults = new RailInfo[](railsLength);
         uint256 resultCount = 0;
 
-        for (uint256 i = 0; i < railsLength; i++) {
+        for (uint256 i = 0; i < railsLength; ) {
             uint256 railId = allRailIds[i];
             Rail storage rail = rails[railId];
 
             // Skip non-existent rails
-            if (rail.from == address(0)) continue;
+            if (rail.from == address(0)) {
+                unchecked {
+                    ++i;
+                }
+                continue;
+            }
 
             // Add rail to our temporary array
             tempResults[resultCount] = RailInfo({
@@ -1645,6 +1653,9 @@ contract Payments is
                 endEpoch: rail.endEpoch
             });
             resultCount++;
+            unchecked {
+                ++i;
+            }
         }
 
         // Create correctly sized final result array
@@ -1652,8 +1663,11 @@ contract Payments is
 
         // Only copy if we have results (avoid unnecessary operations)
         if (resultCount > 0) {
-            for (uint256 i = 0; i < resultCount; i++) {
+            for (uint256 i = 0; i < resultCount; ) {
                 result[i] = tempResults[i];
+                unchecked {
+                    ++i;
+                }
             }
         }
 


aarshkshah1992:
@pali101 Thank you for the PR. Appreciate it but we don't currently see the need for trivial gas savings at the risk of adding unchecked to the contract.

pali101:
Hi @aarshkshah1992, thank you for the review.

I understand the concern around using `unchecked`, but just to clarify: the loops modified here all have clearly defined upper bounds (array lengths), so overflow is provably impossible. This ensures that the use of `unchecked { ++i }` is safe and purely aimed at micro-optimization.

Also worth noting: starting from Solidity 0.8.22, the compiler automatically omits overflow checks in such bounded for loops ([source](https://soliditylang.org/blog/2023/10/25/solidity-0.8.22-release-announcement/)). Since this codebase is using 0.8.20, I applied this manually to reflect future compiler behavior and performance alignment.

This was intended as a first step toward a broader set of iterative optimizations that I had in mind like reducing redundant storage reads, applying bit packing for structs where appropriate, etc. Would those types of improvements be considered useful for the project, or are they also viewed as too minor at this stage?

I'd love to hear your thoughts on whether these optimizations could be valuable contributions moving forward.

Stebalien:
We definitely need to optimize this contract but: correctness first, readability second, performance third.

1. An optimization PR needs to _demonstrate_ a performance benefit proportional to any increases in complexity/risk. The only demonstrated performance change here is a _reduction_ in performance.
2. An optimization PR needs to target performance critical code, methods touched in this PR are not performance critical (view methods).

The first step here would be to submit a PR with some "realistic" benchmarks. Once we have that, we can start figuring out what we need to optimize. Also, we'll want to make sure to run those benchmarks with solidity optimizations enabled: there's no point in making an optimization by-hand that the compiler would make for us.

---

In terms of this increasing gas usage, that's not totally unexpected. We build the [builtin-actors](https://github.com/filecoin-project/builtin-actors) with checked math because, when we looked into disabling checked math after a period of testing, we found that doing so actually decreased performance (likely because the extra checks actually helped the optimizer). Checked math is slow when you're doing a lot of math in a tight loop but smart contracts are 99% business logic. Most of the gas usage is going to come from storage operations.

pali101:
Thanks for the detailed feedback, @Stebalien. I completely agree with your hierarchy: correctness > readability > performance. 

I also realized that, per our foundry.toml, we’re actually compiling with Solidity 0.8.25 which already auto-omits overflow checks in bounded loops, so my manual use of unchecked here is redundant under our current setup.

I’m now working on caching optimizations and have already achieved over 10,000 gas savings. I'll publish the full benchmark results in a follow-up PR.

Which loops or functions would you recommend I prioritize next?

Stebalien:
The most important methods to optimize are:

1. `settleRail` 
2. `modifyRailPayment` and `modifyRailLockup`
3. `createRail`
4. `terminateRail`
5. `settleTerminatedRailWithoutArbitration`

In that order.

But first we'll likely need better benchmarks.

Also note: we'll be using this contract on Filecoin which doesn't use the EVM's gas model (the EVM is run inside the FVM and uses the FVM's gas model). General optimizations (hoisting loads outside of loops, reducing instruction count, etc.) will often carry over, but micro-optimizations targeting nuances in the EVM gas model will not. You can use https://github.com/filecoin-project/ref-fvm/tree/master/tools/fvm-bench benchmark the contract with FVM's gas model, but that's nearly as easy as benchmarking EVM gas.

PR #86, opened 2025-05-09
Optimize the rate-change queue

Stebalien:
This stores the rate-change queue as a single dynamic array instead of a mapping. Given how both the EVM and FEVM work, there's no extra cost for sparse arrays but there is for mappings (especially in FEVM).

From EVM gas benchmarking, this doesn't appear to make much of a difference (and I'm really not clear on why) but it makes the queue take up fewer slots at the very least.

In the FVM, this should ensure that rate changes are stored adjacent to eachother making settlement less expensive.

### DIFF
--- a/src/Payments.sol
+++ b/src/Payments.sol
@@ -1496,9 +1496,6 @@ contract Payments is
             "rate change queue must be empty post full settlement"
         );
 
-        // Clear the rate change queue
-        rail.rateChangeQueue.clear();
-
         rail.token = address(0);
         rail.from = address(0); // This now marks the rail as inactive
         rail.to = address(0);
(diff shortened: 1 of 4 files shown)


src/Payments.sol
```diff
@@ -1496,9 +1496,6 @@ contract Payments is
             "rate change queue must be empty post full settlement"
         );
 
-        // Clear the rate change queue
-        rail.rateChangeQueue.clear();
```
Stebalien, line 1500: The queue is always zero when empty.

aarshkshah1992:
lgtm.

aarshkshah1992 (approved):

PR #95, opened 2025-05-20
fix: centralize zero-rate payment handling

### LINKED ISSUE #83: Don't arbitrate zero-value transfers
We special-case the rate being zero in a few cases like:

https://github.com/FilOzone/fws-payments/blob/07d54a5f1449da85a4d6e392e7dc2c5ed1ffa220/src/Payments.sol#L1162-L1171

But we should probably centralize this logic at the top of `_settleSegment`. Otherwise, e.g., we'll arbitrate zero-rate segments because that "is zero" check only happens on the "rate-change queue is empty" branch.

That would also let us get rid of this special-case: https://github.com/FilOzone/fws-payments/blob/07d54a5f1449da85a4d6e392e7dc2c5ed1ffa220/src/Payments.sol#L963-L978

Chaitu-Tatipamula:
## Description
This PR fixes the inconsistent handling of zero-rate payments. Currently, zero-rate payments are handled in multiple places

## Changes
- Centralized zero-rate payment handling in `_settleSegment`
- Removed redundant zero-rate checks from `settleRail`
- Added early return for zero-rate payments to prevent unnecessary arbitration
- Improved code maintainability by consolidating the logic in one place

## Impact
This change:
- Fixes a potential issue where zero-rate segments might be unnecessarily arbitrated
- Makes the code more maintainable by centralizing zero-rate handling logic

## Related Issues
Closes #83

### DIFF
--- a/src/Payments.sol
+++ b/src/Payments.sol
@@ -984,7 +984,6 @@ contract Payments is
         Account storage payer = accounts[rail.token][rail.from];
 
         // Handle terminated and fully settled rails that are still not finalised
-
         if (isRailTerminated(rail) && rail.settledUpTo >= rail.endEpoch) {
             finalizeTerminatedRail(rail, payer);
             return (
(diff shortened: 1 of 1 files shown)


aarshkshah1992 (approved):

aarshkshah1992:
Thanks !

PR #96, opened 2025-05-20
fix: zero out comission on rail when rail is cleared

### LINKED ISSUE #81: `_zeroOutRail` doesn't clear `commissionRateBps`
I wish there were a better way to zero the rail where this kind of thing couldn't happen.

aarshkshah1992:
Closes #81


aarshkshah1992:
Merging as it's a simple pretty self-evident change.

PR #99, opened 2025-05-23
feat: implement getAccountInfo view function

### LINKED ISSUE #78: Add a method to get account information
We need a way to get account information. This should include:

1. `fundedUntil` -- at what epoch would the account go into debt given the current lockup rate, balance, etc.
2. `totalBalance`.
3. `availableBalance`.
4. `lockupRate`.

Chaitu-Tatipamula:
Added a view function to get account information including:
- fundedUntil: epoch when account would go into debt
- totalBalance: total account balance
- availableBalance: balance minus current lockup
- lockupRate: current lockup rate per epoch

The function simulates account lockup settlement without modifying state,

Closes #78

### DIFF
--- a/src/Payments.sol
+++ b/src/Payments.sol
@@ -1692,39 +1692,30 @@ contract Payments is
         uint256 elapsedTime = currentEpoch - account.lockupLastSettledAt;
         uint256 simulatedLockupCurrent = account.lockupCurrent;
 
-        if (elapsedTime <= 0) {
-            // If no elapsed time, current lockup is the same as account.lockupCurrent
+        // Early return for simple cases: no elapsed time or no lockup rate
+        if (elapsedTime <= 0 || account.lockupRate == 0) {
             availableFunds = account.funds > simulatedLockupCurrent ? 
                 account.funds - simulatedLockupCurrent : 0;
             
-            // Calculate fundedUntil based on lockupRate
-            if (account.lockupRate == 0) {
-                fundedUntilEpoch = type(uint256).max;
-            } else if (availableFunds == 0) {
-                fundedUntilEpoch = account.lockupLastSettledAt;
-            } else {
-                uint256 epochsUntilDebt = availableFunds / account.lockupRate;
-                fundedUntilEpoch = currentEpoch + epochsUntilDebt;
-            }
+            // If no lockup rate, account never goes into debt
+            fundedUntilEpoch = account.lockupRate == 0 ? type(uint256).max : availableFunds == 0 ? account.lockupLastSettledAt : currentEpoch + (availableFunds / account.lockupRate);
             
             return (fundedUntilEpoch, currentFunds, availableFunds, currentLockupRate);
         }
 
         // Handle case where we need to calculate additional lockup
-        if (account.lockupRate > 0) {
-            uint256 additionalLockup = account.lockupRate * elapsedTime;
-            
-            // If we have sufficient funds to cover the additional lockup
-            if (account.funds >= account.lockupCurrent + additionalLockup) {
-                simulatedLockupCurrent = account.lockupCurrent + additionalLockup;
-            } else {
-                // Calculate partial settlement
-                uint256 availableForLockup = account.funds - account.lockupCurrent;
-                if (availableForLockup > 0) {
-                    // Calculate how many epochs we can fund with available funds
-                    uint256 fractionalEpochs = availableForLockup / account.lockupRate;
-                    simulatedLockupCurrent = account.lockupCurrent + (account.lockupRate * fractionalEpochs);
-                }
+        uint256 additionalLockup = account.lockupRate * elapsedTime;
+        
+        // If we have sufficient funds to cover the additional lockup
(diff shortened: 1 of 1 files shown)


src/Payments.sol
```diff
@@ -1665,6 +1665,69 @@ contract Payments is
...
+
+        if (account.lockupRate == 0) {
+            // If no lockup rate, account is funded indefinitely
+            fundedUntil = type(uint256).max;
+        } else {
+            // Calculate how many epochs the available funds can cover
+            uint256 availableFunds = account.funds - simulatedLockupCurrent;
+            if (availableFunds == 0) {
+                // If no available funds, account is already in debt
+                fundedUntil = block.number;
```
silent-cipher, line 1723: I think we can show more appropriate value for `fundedUntil` in case of debt
```suggestion
                uint256 fractionalEpochs = ( simulatedLockupCurrent - account.lockupCurrent ) / account.lockupRate;
                fundedUntil = account.lockupLastSettledAt + fractionalEpochs;
```

Chaitu-Tatipamula: yes this would be great!

aarshkshah1992:
@Chaitu-Tatipamula 

Please can you rebase https://github.com/FilOzone/filecoin-services-payments/pull/99/files on master and add tests as well ? I don't see any tests in the PR. Some solid tests that assert that the account state we see is the expected state after doing some lockups, rate changes and settlement would be great

src/Payments.sol
```diff
@@ -1664,6 +1664,70 @@ contract Payments is
...
+        Account storage account = accounts[token][owner];
+        
+        totalBalance = account.funds;
+        lockupRate = account.lockupRate;
+
+        uint256 currentEpoch = block.number;
+        uint256 elapsedTime = currentEpoch - account.lockupLastSettledAt;
+        uint256 simulatedLockupCurrent = account.lockupCurrent;
+
+        if (elapsedTime > 0 && account.lockupRate > 0) {
```
aarshkshah1992, line 1695: @Chaitu-Tatipamula Would be best to have a seperate block for `elapsedTime  <= 0 OR  account.lockupRate > 0==0` and return right away the correct values for that case. There is some complexity here because of trying to understand what happens if this ` if (elapsedTime > 0 && account.lockupRate > 0)` block is NOT hit.

src/Payments.sol
```diff
@@ -1664,6 +1664,70 @@ contract Payments is
...
+    /**
+     * @notice Gets information about an account - when it would go into debt, total balance, available balance, and lockup rate.
+     * @param token The token address to get account info for.
+     * @param owner The address of the account owner.
+     * @return fundedUntil The epoch at which the account would go into debt given current lockup rate and balance.
+     * @return totalBalance The total balance of the account.
+     * @return availableBalance The available balance (total balance minus current lockup).
+     * @return lockupRate The current lockup rate per epoch.
+     */
+    function getAccountInfo(
```
aarshkshah1992, line 1677: this should be `simulateAccountSettlement` or `getAccountInfoIfSettled`

aarshkshah1992: Please also change all returned variable names accordingly. Especially `totalBalance ` and `availableBalance `.

aarshkshah1992: @Chaitu-Tatipamula Yeah makes sense.

src/Payments.sol
```diff
@@ -1664,6 +1664,85 @@ contract Payments is
...
+                account.funds - simulatedLockupCurrent : 0;
+            
+            // Calculate fundedUntil based on lockupRate
+            if (account.lockupRate == 0) {
+                fundedUntilEpoch = type(uint256).max;
+            } else if (availableFunds == 0) {
+                fundedUntilEpoch = account.lockupLastSettledAt;
+            } else {
+                uint256 epochsUntilDebt = availableFunds / account.lockupRate;
+                fundedUntilEpoch = currentEpoch + epochsUntilDebt;
```
aarshkshah1992, line 1707: this makes sense.

src/Payments.sol
```diff
@@ -1664,6 +1664,85 @@ contract Payments is
...
+                    simulatedLockupCurrent = account.lockupCurrent + (account.lockupRate * fractionalEpochs);
+                }
+            }
+        }
+
+        // Calculate available balance and fundedUntilEpoch
+        availableFunds = account.funds > simulatedLockupCurrent ? 
+            account.funds - simulatedLockupCurrent : 0;
+
+        if (account.lockupRate == 0) {
```
aarshkshah1992, line 1735: @Chaitu-Tatipamula Most of the code in this ` if (account.lockupRate == 0) {` block is similar to the `   if (elapsedTime <= 0) {` block. Let's combine both the blocks.

Chaitu-Tatipamula:
```
if (account.lockupRate == 0 || elapsedTime <= 0) {
            fundedUntilEpoch = type(uint256).max;
        } else if (availableFunds == 0) {
```

this should work then

ZenGround0:
Please note the simplified implementation here #154

Issue #114, opened 2025-06-06
Payments contract: commission recipient architecture enhancement


jennijuju:
## Problem Statement

While reviewing FilOzone/filecoin-services#37 and implementing FilOzone/filecoin-services#38, I found that the current Payments contract design creates accounting challenges for services that collect commission fees. When a service contract acts as the operator (commission recipient), it leads to mixed fund scenarios:

1. **Mixed Funds**: Service contracts often hold operational funds (gas costs, maintenance, reserves, dao funding and etc) alongside commission earnings
2. **Accounting Complexity**: No clear separation between "earned commissions" and "operational deposits" 
3. **Risk of Error**: Potential to accidentally withdraw operational funds when collecting earned commissions
4. **Lack of Transparency**: Difficult to audit commission earnings vs. operational expenses across multiple services

## Current Architecture Limitation

In the existing `createRail` function, the operator (commission recipient) is automatically set to `msg.sender`:

```solidity
function createRail(
    address token,
    address from,
    address to, 
    address arbiter,
    uint256 commissionRateBps
) external returns (uint256) {
    address operator = msg.sender; // ❌ Operator is forced to be the calling contract
    // ...
}
```

This means:
- Service contracts must receive commissions in their own account
- No way to separate commission earnings from other funds in the service contracts 
- All services using Payments face the same accounting challenges

### Option 1: Service-Level Commission Tracking *(Workaround)*

**Approach**: Each service implements internal tracking of commission earnings
- Services add commission tracking mappings to their contracts
- Track incremental commission earnings separately from operational funds
- Provide collection functions that only withdraw tracked commission amounts

this approach doesnt require any payment contract update, but I am less of a fan of this approach cuz
- ❌ Duplicated tracking logic across all services
- ❌ More complex state management for each service
- ❌ Potential for tracking bugs or edge cases in each implementation
- ❌ Manual collection step required by each service
- ❌ No standardized approach across services

### Option 2: Enhanced Payments Contract *(Recommended)*

**Approach**: Modify the Payments contract to accept a commission recipient address
- Add `commissionRecipient` parameter to `createRail()` function
- During settlement, credit commissions directly to the specified recipient
- Decouple commission destination from the calling service contract

**Current `createRail` signature**:
```solidity
function createRail(
    address token,
    address from,
    address to,
    address arbiter,
    uint256 commissionRateBps
) external returns (uint256)
```

**Proposed enhanced signature**:
```solidity
function createRail(
    address token,
[...]

PR #118, opened 2025-06-11
feat:commission recipient architecture enhancement

SarveshLimaye:
Fixes : #114 

### Description 

This pr adds all the enhancements mentioned in : #114 

### Changes - 

- Added commissionRecipient parameter to createRail() function
- If serviceFeeRecipient  is provided, commissions go there during the settlement
- If serviceFeeRecipient  is address(0), operator recieves the commission

### DIFF
--- a/src/Payments.sol
+++ b/src/Payments.sol
@@ -69,6 +69,7 @@ contract Payments is
         uint256 endEpoch; // Final epoch up to which the rail can be settled (0 if not terminated)
         // Operator commission rate in basis points (e.g., 100 BPS = 1%)
         uint256 commissionRateBps;
+        address serviceFeeRecipient; // address to collect operator comission
     }
 
     struct OperatorApproval {
@@ -103,6 +104,7 @@ contract Payments is
         uint256 endEpoch;
         // Operator commission rate in basis points (e.g., 100 BPS = 1%)
         uint256 commissionRateBps;
+        address serviceFeeRecipient; // address to collect operator commission
     }
 
     // token => client => operator => Approval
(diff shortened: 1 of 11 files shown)


src/Payments.sol
```diff
@@ -510,6 +515,9 @@ contract Payments is
         rail.settledUpTo = block.number;
         rail.endEpoch = 0;
         rail.commissionRateBps = commissionRateBps;
+        rail.serviceFeeRecipient = serviceFeeRecipient == address(0)
```
aarshkshah1992, line 518: @SarveshLimaye Let's not do this. Let's revert `createRail ` if commission is non zero but `serviceFeeRecipient ` is 0 with a meaningful error message that the service contract dosen't end up doing something unexpected wrt collecting the comission.

SarveshLimaye: @aarshkshah1992 Sure makes sense. Implemented this in the recent commit

rjan90:
@aarshkshah1992 With the requested changes implemented in the latest commit, can this be merged now?

src/Payments.sol
```diff
@@ -832,8 +843,17 @@ contract Payments is
 
         // Credit operator (if commission exists)
         if (operatorCommission > 0) {
-            Account storage operatorAccount = accounts[token][operator];
-            operatorAccount.funds += operatorCommission;
+            // Ensure the contract has enough funds to pay the operator commission
+            require(
+              IERC20(token).balanceOf(address(this)) >= operatorCommission,
```
aarshkshah1992, line 848: @SarveshLimaye I don't think understand why we do this now. We want to keep the funds for the commission in the payments contract account of this `serviceFeeRecipient `. Why are we doing an ERC-20 transfer ? Can you explain this decision in detail ?

aarshkshah1992: @SarveshLimaye Yes, let's keep the funds in the Payments contract for now. Owners of the account can always call `withdraw` if they wanna move funds out of the contract.

aarshkshah1992:
@SarveshLimaye Merged ! Thank you !

PR #125, opened 2025-06-17
fix: Optimistic Lockup Modification for Rail Updates

### LINKED ISSUE #116: Modification of rail lockup prevented unless same epoch as settlement
Modifying a rail lockup requires `isAccountLockupFullySettled`, but this is a bit confusingly named. It checks that the last settlement epoch is this epoch, but that's much stricter than whether the fixed lockup is settled. The error message "insufficient funds to cover the current lockup" is not necessarily correct.

https://github.com/FilOzone/filecoin-services-payments/blob/182b414d31005e6002c43eddfe7c4db344d1cdce/src/Payments.sol#L592-L603

Consider proceeding to optimistically change the lockup, and revert if necessary after calculating the actual funds needed.

Chaitu-Tatipamula:
## Problem
The previous implementation of `modifyNonTerminatedRailLockup` had overly restrictive validation that prevented legitimate lockup modifications. The function used `isAccountLockupFullySettled()` which only checked if the account lockup was settled up to the current epoch, but this was much stricter than what was actually needed for safe lockup modifications.

The error message "insufficient funds to cover the current lockup" was misleading because the function didn't actually verify sufficient funds - it only checked settlement status.

## Solution
Implemented an optimistic approach for lockup modifications:

1. **Removed restrictive pre-validation**: Eliminated the overly strict `isAccountLockupFullySettled` check that prevented legitimate modifications
2. **Optimistic lockup calculation**: Allow the operation to proceed and calculate the actual lockup difference
3. **Post-execution validation**: Rely on the `settleAccountLockupBeforeAndAfterForRail` modifier to validate the operation after execution

### `modifyNonTerminatedRailLockup` function:
- Removed restrictive `isAccountLockupFullySettled` checks
- Implemented optimistic lockup difference calculation

## Related Issues
Fixes #116

### DIFF
--- a/src/Payments.sol
+++ b/src/Payments.sol
@@ -628,19 +628,6 @@ contract Payments is
     ) internal {
         Account storage payer = accounts[rail.token][rail.from];
 
-        // Don't allow changing the lockup period or increasing the fixed lockup unless the payer's
-        // account is fully settled.
-        if (!isAccountLockupFullySettled(payer)) {
-            require(
-                period == rail.lockupPeriod,
-                "cannot change the lockup period: insufficient funds to cover the current lockup"
-            );
-            require(
-                lockupFixed <= rail.lockupFixed,
-                "cannot increase the fixed lockup: insufficient funds to cover the current lockup"
-            );
-        }
-
         // Get operator approval
         OperatorApproval storage operatorApproval = operatorApprovals[
             rail.token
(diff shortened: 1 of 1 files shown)


aarshkshah1992:
@Chaitu-Tatipamula Closing this as per the discussion. Sorry about that !

PR #126, opened 2025-06-17
Migrate ReentrancyGuard with upgradeable version in Payments contract

pali101:
## Summary

This PR updates the Payments contract to use `ReentrancyGuardUpgradeable` instead of the regular `ReentrancyGuard`, which is necessary for compatibility with the proxy-based upgradeable pattern.

Other OpenZeppelin imports (`SafeERC20`, `IERC20`, `IERC20Permit`, and `Strings`) remain unchanged as they are stateless/interfaces and safe to use in upgradeable contracts.


ZenGround0 (approved):
Very nice

PR #142, opened 2025-06-20
fix: use 'to' parameter instead of msg.sender in depositWithPermit

rvagg:
This change makes it compatible with USDFC.

I have to admit to not having a super solid theory as to why this works and using `msg.sender` doesn't, but when I switch it out I can make it work, but with `msg.sender` I get a [`DebtToken: invalid signature`](https://github.com/Secured-Finance/stablecoin-contracts/blob/develop/contracts/DebtToken.sol#L182-L211) failure.

This change seems to fix the issue by using the `to` parameter instead of `msg.sender` in the permit call. Since `to` is already the owner who signed the permit, this is semantically correct. BUT since they should be the same value when we make this I don't have a good explanation as to why this makes a difference, but it does.

Deploying a contract with this change on calibnet lets me do a permit sign + single transaction to get setup with USDFC.

@pali101 perhaps you have some insight or ideas here?

### DIFF
--- a/src/Payments.sol
+++ b/src/Payments.sol
@@ -142,7 +142,7 @@ contract Payments is
     }
 
     // Events
-    event DepositWithPermit(address indexed token, address indexed from, address indexed to, uint256 amount);
+    event DepositWithPermit(address indexed token, address indexed account, uint256 amount);
 
     /// @custom:oz-upgrades-unsafe-allow constructor
     constructor() {
@@ -419,7 +419,7 @@ contract Payments is
     /**
     * @notice Deposits tokens using permit (EIP-2612) approval in a single transaction.
     * @param token The ERC20 token address to deposit.
-    * @param to The address whose account will be credited.
+    * @param to The address whose account will be credited (must be the permit signer).
     * @param amount The amount of tokens to deposit.
     * @param deadline Permit deadline (timestamp).
     * @param v,r,s Permit signature.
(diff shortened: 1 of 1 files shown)


rvagg:
This fails the `testDepositWithPermitToAnotherUser` test .. of course, so probably not quite right yet if we want to support that use case.

pali101:
Hi @rvagg 

If we want to ensure that only token owner  (`to`)  calls `depositWithPermit` then modified code works. However, if we plan to support gasless transactions or enable a relayer in the future, we will need to refactor this function.

Also, I realized that I could have chosen better variable names to minimize confusion around who signs, who is credited, and who submits the transaction. Let me know if you want help updating the code.

rvagg:
@pali101 ok, I'll take your word for it on what we need to do to make this maximally compatible with future options, but right now I'm just solving for _making it work_ at all and I don't really mind what the solution is. All I can tell though is that `msg.sender` isn't working, which doesn't make sense to me, but using an explicit `to` does work.

I think maybe any refactoring of this code should come with some basic integration tests to make sure it's actually usable, with USDFC at least. That's where I'm getting stuck. This code works in theory, and with unit tests, but I can't make it work in practice with the contract that we want to support, so it's essentially dead code for us.

aarshkshah1992:
lgtm in the interest of making this work for now.

aarshkshah1992:
I have removed the failing test for now as supporting the "gasless transaction/different relayer" functionality isn't a high priority right now.

Filed https://github.com/FilOzone/filecoin-services-payments/issues/146 to track this and discuss/implement if we really want to support this functionality.

PR #149, opened 2025-06-24
WIP: Payments contract should compile without via-ir being set to true in foundry.toml

pali101:
Refactored the Payments contract to reduce call stack depth and remove the requirement for `via-ir` in `foundry.toml`.

## Breaking Changes

To address stack depth limits, I had to change function signatures of few external/public function (e.g., `settleRail`, `settleTerminatedRailWithoutArbitration` etc). Key changes:
- Grouped related function parameters into structs (e.g., `PermitParams`).
- Replaced multi-variable returns with result structs.
- Moved some logic to internal helpers.

## Remaining work
Update tests to use the new function signatures and verify full functionality.

### DIFF
--- a/foundry.toml
+++ b/foundry.toml
@@ -6,7 +6,6 @@ out = 'out'
 libs = ['lib']
 cache_path = 'cache'
 solc = "0.8.25"
-via_ir = true
 
 # For dependencies
 remappings = [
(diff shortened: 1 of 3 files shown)


aarshkshah1992:
@pali101 Let's get back to this after the M2 milestone. Thanks for your hard work so far.

BigLep:
@wjmelements : is this still worth doing, or should we close this?

PR #153, opened 2025-06-24
feat: function to introspect if rail is in debt

### LINKED ISSUE #140: Implement `isRailInDebt` function
- `isRailInDebt` function would allow operators to check and have a terminate function which payee can run if rail goes to debt to get the locked amount of SP.
- Maybe we can implement this here only?

aarshkshah1992:
Closes #140 .

### DIFF
--- a/src/Payments.sol
+++ b/src/Payments.sol
@@ -313,6 +313,28 @@ contract Payments is
             });
     }
 
+    /// @notice Checks if a rail is in debt by settling the payer's account lockup and checking if it's behind.
+    /// @param railId The ID of the rail to check.
+    /// @return True if the rail is in debt (payer's lockup cannot be fully settled to current epoch), false otherwise.
+    function isRailInDebt(
+        uint256 railId
+    ) external validateRailActive(railId) nonReentrant returns (bool) {
+        Rail storage rail = rails[railId];
+        
+        // Zero rate rails with empty rate change queues never contribute to debt
+        if (rail.paymentRate == 0 && rail.rateChangeQueue.isEmpty()) {
+            return false;
+        }
+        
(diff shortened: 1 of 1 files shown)


src/Payments.sol
```diff
@@ -313,6 +313,28 @@ contract Payments is
...
+        
+        // Zero rate rails with empty rate change queues never contribute to debt
+        if (rail.paymentRate == 0 && rail.rateChangeQueue.isEmpty()) {
+            return false;
+        }
+        
+        Account storage payerAccount = accounts[rail.token][rail.from];
+        
+        // Settle the account lockup and get the epoch it was settled to
+        uint256 settledUpToEpoch = settleAccountLockup(payerAccount);
```
ZenGround0, line 332: If we do want to keep this PR then I don't think we want to actually settle, I think we want to follow the approach of #99.

I think we're better off either just leaving the solution in #99 for the operator to call, or if we really want an onchain method using rail id we can replace all this with 

```solidity
(uint256 fundedUntilEpoch,,,) =  getAccountInfoIfSettled(rail.token, rail.from);
return fundedUntilEpoch < currentEpoch
```

ZenGround0 (changes requested):
Not clear if we just want to close this and allow operator to directly call `getAccountInfoIfSettled` or if we want a wrapper around the railId.  If its the first case we can just close this down. If its the second we should wrap a call to `getAccountInfoIfSettled`

aarshkshah1992:
Closing this in favour of  https://github.com/FilOzone/filecoin-services-payments/pull/99 and https://github.com/FilOzone/filecoin-services-payments/pull/154

PR #154, opened 2025-06-24
Simplify #99

ZenGround0:
PR #99 is not too bad but I have a guess that it can be simplified by such a degree that it is worth spending some time to enforce code quality.  

I'm almost certainly missing some cases so I'll leave this in draft until I get tests passing.


src/Payments.sol
```diff
@@ -1417,48 +1417,16 @@ contract Payments is Initializable, UUPSUpgradeable, OwnableUpgradeable, Reentra
...
 
-        // Handle case where we need to calculate additional lockup
-        uint256 additionalLockup = account.lockupRate * elapsedTime;
+        fundedUntilEpoch = account.lockupRate == 0
+            ? type(uint256).max
+            : account.lockupLastSettledAt + (account.funds - account.lockupCurrent) / account.lockupRate;
+        uint256 simulatedSettledAt = fundedUntilEpoch >= currentEpoch ? currentEpoch : fundedUntilEpoch;
+        uint256 simulatedLockupCurrent =
+            account.lockupCurrent + account.lockupRate * (simulatedSettledAt - account.lockupLastSettledAt);
+        availableFunds = account.funds - simulatedLockupCurrent;
```
aarshkshah1992, line 1427: just rename this to `simulatedAvailableFunds `

aarshkshah1992:
This is a great refactor ! Thank you so much !

PR #159, opened 2025-06-26
fix: do not allow rate modification or one time payments for finalised rails

### LINKED ISSUE #136: bug: do not enqueue a rate change for finalised queues
See `modifyPaymentForTerminatedRailBeyondLastEpoch`

It should not exist at all as it makes the rate change queue hard to reason about. Instead, simply reject all requests to change the rate of a rail that is beyond it's `endEpoch.`

aarshkshah1992:
Closes #136


ZenGround0 (approved):
LGTM

PR #160, opened 2025-06-26
Require network fee to be paid when settling a rail

Kubuxu:
Introduce a network of 130000attoFIL when settling a rail.

Still investigating work required on services and SDK.
Also test for the fee itself also incoming.

### DIFF
--- a/src/Payments.sol
+++ b/src/Payments.sol
@@ -10,6 +10,7 @@ import "@openzeppelin/contracts-upgradeable/access/OwnableUpgradeable.sol";
 import "./RateChangeQueue.sol";
 import "@openzeppelin/contracts/utils/Strings.sol";
 import "@openzeppelin/contracts/token/ERC20/extensions/IERC20Permit.sol";
+import {console} from "forge-std/console.sol";
 
 interface IValidator {
     struct ValidationResult {
@@ -41,6 +42,8 @@ contract Payments is Initializable, UUPSUpgradeable, OwnableUpgradeable, Reentra
     uint256 public constant COMMISSION_MAX_BPS = 10000;
 
     uint256 public constant PAYMENT_FEE_BPS = 10; //(0.1 % fee)
+
+    uint256 public constant NETWORK_FEE = 1300000 gwei; // equivalent to 130000 nFIL
 
     // Events
     event AccountLockupSettled(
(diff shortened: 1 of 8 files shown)


src/Payments.sol
```diff
@@ -789,6 +794,15 @@ contract Payments is Initializable, UUPSUpgradeable, OwnableUpgradeable, Reentra
         return settleRailInternal(railId, maxSettleEpoch, true);
     }
 
+    function burn(uint256 _amount) internal {
+        require(msg.value >= _amount, "not enough native token to burn");
+        // f099 burn address
+        (bool success,) = address(0xff00000000000000000000000000000000000063).call{value: _amount}("");
```
Kubuxu, line 800: I need to check if this will work

aarshkshah1992: @Kubuxu This is how we burn in the `PDPVerifier`:

https://github.com/FilOzone/pdp/blob/d051560b953072942ea5fc8ae8d31b9b0c2e7021/src/PDPVerifier.sol#L30

aarshkshah1992:
@Kubuxu Is this ready for a review now ? Are you done with your changes ?

src/Payments.sol
```diff
@@ -912,7 +918,17 @@ contract Payments is Initializable, UUPSUpgradeable, OwnableUpgradeable, Reentra
         return settleRailInternal(railId, maxSettleEpoch, true);
     }
 
+    function burn(uint256 _amount) internal {
+        require(msg.value >= _amount, "insufficient transfer of native token to burn");
+        // f099 burn address
+        (bool success,) = address(0xff00000000000000000000000000000000000063).call{value: _amount}("");
+        require(success, "native token burn failed");
+
+        // TODO(Kubuxu): maybe refund? but also re-entrancy
```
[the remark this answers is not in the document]
aarshkshah1992: Maybe park it in their account so they withdraw whenever they want.

Kubuxu:
> You need to run forge fmt before pushing to the linter CI is green.

It was green but the merge from master broke it, will fix.

Kubuxu:
@aarshkshah1992 updated

aarshkshah1992 (approved):
lgtm

rvagg:
@Kubuxu I'm just looking at this now from the SDK and noting that settleTerminatedRailWithoutValidation doesn't burn, should it? Is that a way to get around the burn - get it terminated and then settle that way?

Kubuxu:
Seems like I missed that path, it should require fee as well.

PR #161, opened 2025-06-30
feat: zero out old payment fee mechanism that charged a percentage of settlement

aarshkshah1992:
Now that https://github.com/FilOzone/filecoin-services-payments/pull/160 has landed which charges a fixed flat fee per settlement, this PR  sets the legacy payment fee which would be a % of the settlement amount to 0.


aarshkshah1992:
@Kubuxu Removed all the code that takes a % payment fee and the corresponding tests.

Issue #148, opened 2025-06-23
Payment Contract Specification - Audit Readiness


rjan90:
## Summary
Enhance the existing FWS Payments Contract specification to ensure it meets audit requirements. The current README.md provides a solid foundation, but additional technical details and formal specifications are needed for a comprehensive security audit.

## Current State
- ✅ High-level contract overview and key concepts
- ✅ Core function documentation with parameters and requirements
- ✅ Worked example demonstrating typical usage flows
- ✅ Emergency scenarios and escape hatches
- ✅ Basic deployment information

### 1. Formal Mathematical Specifications
- [ ] **Lockup Calculations**: Provide formulas for:
  - `lockupCurrent` computation: `rate × lockupPeriod + fixedLockup`
  - Account funding threshold calculations
  - Settlement amount calculations with epoch-based prorating
- [ ] **State Transitions**: definitions of state changes during:
  - Rail creation and modification
  - Settlement operations
  - Termination scenarios

### 2. Complete State Machine Documentation
- [ ] **Account States**: Document all possible account states and valid transitions
- [ ] **Rail Lifecycle**: Detailed state diagram showing rail states (Active, Terminated, Settled)
- [ ] **Operator State Management**: How operator allowances interact with rail states
- [ ] **Settlement State Tracking**: Precise definition of settlement epochs and partial settlement states

### 3. Security Model and Threat Analysis
- [ ] **Trust Assumptions**: Clearly define trust relationships between:
  - Clients and Operators
  - Operators and Service Providers
  - All parties and Arbiters
- [ ] **Attack Vectors**: Document potential attack scenarios and mitigations:
  - Operator misbehavior
  - Client underfunding attacks
  - Arbiter manipulation
  - Reentrancy considerations
- [ ] **Economic Incentives**: Analysis of economic attack vectors and game theory

### 4. Contract Invariants
- [ ] **System Invariants**: Formal specification of properties that must always hold:
  - Total locked funds ≤ total deposited funds
  - Settlement epochs are monotonic
  - Operator allowances are respected
- [ ] **Rail Invariants**: Properties specific to individual rails:
  - Settlement cannot exceed what's owed
  - Terminated rails cannot increase rates
  - Lockup periods are immutable for terminated rails

### 5. Integration Specifications
- [ ] **Arbiter Interface**: Detailed specification of `IArbiter` interface:
  - Expected behavior for different return values
  - Error handling requirements
  - Performance considerations
[...]

PR #163, opened 2025-07-01
Fixup README

ZenGround0:
A few small README fixups in preparation for #148

### Rename arbiter to validator 

Note I kept a small number of references to validation as validation/arbitration.  Eva wanted the rename because in the common case there is no arbitration.  But in the generic case arbitration is a good description of what is happening and theres no need to scrub the word from the README entirely.

### Add functions to interface 
* getAccountInfoIfSettled


aarshkshah1992 (approved):
Thanks !

PR #166, opened 2025-07-02
feat(errors): add centralized custom error library for protocol

### LINKED ISSUE #156: Contract Size is close to the 24KB limit
See https://filecoinproject.slack.com/archives/C07CGTXHHT4/p1750762395475799

```
We might need to refactor the Payments contract soon, as we’re close to hitting the 24 KB contract size limit.
With the optimizer enabled (optimizer_runs = 20,000) to keep per-call costs low, the runtime size is already 23,384 bytes, 
leaving a margin of just 1,192 bytes.

If we increase optimizer_runs to 50,000 (to further reduce gas for high-traffic use cases), the runtime size actually exceeds 
the 24 KB limit. 

Suggested solution
To reclaim contract size, we can remove string error messages and replace them with custom errors. 
For returning human readable errors, we can write a function in sdk that return error name by matching error selector from 
all error in abi (e.g., decodeContractError)
```

Thanks a lot to @pali101 for reporting this !

pali101:
## Summary

This PR refactors the Payments contract and its test suite to use a centralized custom error library (`Errors.sol`) instead of string-based revert messages.

## Key changes

- Adds a protocol-level `Errors.sol` with descriptive custom errors for core revert scenarios.
- Refactors most `require` and `revert` statements in `Payments.sol` to use custom errors
- Updates test suite to check for revert selectors and error parameters rather than error strings.

## Results

| Metric                  | Before (Strings) | After (Custom Errors) | Change       |
|-------------------------|------------------|-----------------------|--------------|
| Deployment Size (bytes) | 22,118           | 19,553                | **-2,565**   |
| Deployment Cost (gas)   | 4,838,144        | 4,284,257             | **-553,887** |

_These results are **without** the optimizer turned on. I'll share the optimized results later._

Closes #156

### DIFF
--- a/src/Errors.sol
+++ b/src/Errors.sol
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: MIT
-pragma solidity ^0.8.20;
+pragma solidity ^0.8.27;
 
 /// @title Errors
 /// @notice Centralized library for custom error definitions across the protocol
--- a/src/Payments.sol
+++ b/src/Payments.sol
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: MIT
-pragma solidity ^0.8.20;
+pragma solidity ^0.8.27;
 
 import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
--- a/foundry.toml
+++ b/foundry.toml
@@ -5,7 +5,7 @@ script = 'script'
 out = 'out'
 libs = ['lib']
 cache_path = 'cache'
-solc = "0.8.25"
+solc = "0.8.30"
 via_ir = true
 
 # For dependencies
--- a/src/ERC1967Proxy.sol
+++ b/src/ERC1967Proxy.sol
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: MIT
 
-pragma solidity ^0.8.20;
+pragma solidity ^0.8.27;
 
 import {ERC1967Proxy} from "@openzeppelin/contracts/proxy/ERC1967/ERC1967Proxy.sol";
 
--- a/src/RateChangeQueue.sol
+++ b/src/RateChangeQueue.sol
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: MIT
-pragma solidity ^0.8.20;
+pragma solidity ^0.8.27;
 
 library RateChangeQueue {
     struct RateChange {
--- a/test/AccountLockupSettlement.t.sol
+++ b/test/AccountLockupSettlement.t.sol
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: MIT
-pragma solidity ^0.8.20;
+pragma solidity ^0.8.27;
 
 import {Test} from "forge-std/Test.sol";
 import {Payments} from "../src/Payments.sol";
--- a/test/AccountManagement.t.sol
+++ b/test/AccountManagement.t.sol
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: MIT
-pragma solidity ^0.8.20;
+pragma solidity ^0.8.27;
 
 import {Test} from "forge-std/Test.sol";
 import {Payments} from "../src/Payments.sol";
--- a/test/DepositWithPermitAndOperatorApproval.t.sol
+++ b/test/DepositWithPermitAndOperatorApproval.t.sol
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: MIT
-pragma solidity ^0.8.20;
+pragma solidity ^0.8.27;
 
 import {Test} from "forge-std/Test.sol";
 import {Payments} from "../src/Payments.sol";
--- a/test/ERC1967Proxy.t.sol
+++ b/test/ERC1967Proxy.t.sol
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: MIT
-pragma solidity ^0.8.20;
+pragma solidity ^0.8.27;
 
 import {Test} from "forge-std/Test.sol";
 import {Payments} from "../src/Payments.sol";
--- a/test/FeeOnTransferVulnerability.t.sol
+++ b/test/FeeOnTransferVulnerability.t.sol
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: MIT
-pragma solidity ^0.8.20;
+pragma solidity ^0.8.27;
 
 import {Test, console} from "forge-std/Test.sol";
 import {Payments} from "../src/Payments.sol";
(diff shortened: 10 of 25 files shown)


rjan90:
@pali101 looks like there is a couple of conflicts that needs to be resolved here - could you sort those out, and then we will get it reviewed

src/Payments.sol
```diff
@@ -197,42 +198,54 @@ contract Payments is Initializable, UUPSUpgradeable, OwnableUpgradeable, Reentra
...
+        if (rails[railId].endEpoch != 0) {
+            revert Errors.RailAlreadyTerminated(railId);
+        }
         _;
     }
 
     modifier validateRailTerminated(uint256 railId) {
-        require(isRailTerminated(rails[railId]), "can only be used on terminated rails");
+        if (!isRailTerminated(rails[railId])) {
+            revert Errors.RailNotTerminated();
```
Kubuxu, line 240: should this also carry `railId` same as `RailAlreadyTerminated`

Kubuxu (approved):
other than one nit, LGTM

src/Payments.sol
pali101: I initially considered adding `railId` for `RailNotTerminated` and `RailDoesNotExist` but this would require us to either modify the `Rail` to include `railId` (since the current code passes the entire rail object to functions like  `remainingEpochsForTerminatedRail`, `maxSettlementEpochForTerminatedRail`, `isRailTerminated` etc) or start passing `railId` as an additional parameter to those functions. 

I'd prefer not to change the function signatures if possible. If necessary, I can update the Rail struct to include `railId` so it can be included in the error. Let me know if you think that approach makes sense, or if you have a preferred way to handle this.

src/Errors.sol
```diff
@@ -0,0 +1,295 @@
...
+    /// @notice The rail does not exist
+    error RailDoesNotExist();
+
+    /// @notice The provided address is zero, which is not allowed
+    /// @param varName The name of the variable that was expected to be non-zero
+    error ZeroAddressNotAllowed(string varName);
+
+    /// @notice The rail is inactive
+    /// @param railId The ID of the rail that is inactive
+    error RailInactive(uint256 railId);
```
ZenGround0, line 46: This appears to be only used in exactly the same scenario as `RailInactiveOrSettled`, i.e. the `from` field is address(0).  You should consolidate the two.

src/Payments.sol
```diff
@@ -373,10 +391,9 @@ contract Payments is Initializable, UUPSUpgradeable, OwnableUpgradeable, Reentra
         Account storage payer = accounts[rail.token][rail.from];
 
         // Only client with fully settled lockup or operator can terminate a rail
-        require(
-            (msg.sender == rail.from && isAccountLockupFullySettled(payer)) || msg.sender == rail.operator,
-            "caller is not authorized: must be operator or client with settled lockup"
-        );
+        if (!((msg.sender == rail.from && isAccountLockupFullySettled(payer)) || msg.sender == rail.operator)) {
```
ZenGround0, line 394: nit: I understand we are removing require for expense reasons.  But for big conditions like this I definitely miss the clarity of require asserting a condition that must be true.  Would it add to expenses too much if we made use of a function like 

```solidity
function myRequire(bool pred, error err) internal {
   if (!pred) {
      revert err;
}
```

Any opinions on this @aarshkshah1992  or @Kubuxu ?  If no one else sees this as more clear then I will drop it.

src/Errors.sol
```diff
@@ -0,0 +1,295 @@
...
+    /// @param railId The ID of the rail being settled
+    /// @param maxAllowedEpoch The latest epoch that can be settled (expected, must be >= actual)
+    /// @param attemptedEpoch The epoch up to which settlement was attempted (actual)
+    error CannotSettleFutureEpochs(uint256 railId, uint256 maxAllowedEpoch, uint256 attemptedEpoch);
+
+    /// @notice No progress was made in settlement; settledUpTo did not advance.
+    /// @param railId The ID of the rail
+    /// @param expectedSettledUpTo The expected value for settledUpTo (must be > startEpoch)
+    /// @param actualSettledUpTo The actual value after settlement attempt
+    error NoProgressInSettlement(uint256 railId, uint256 expectedSettledUpTo, uint256 actualSettledUpTo);
```
ZenGround0, line 209: nit: it might be useful to take invariant violations like this and give their errors a common prefix so we can quickly look at errors and know if they represent programmer errors.  i.e. `InvariantViolationNoProgressInSettlement` or `InternalErrNoProgressInSettlement`.  Just an idea.  It would be a decent amount of more work because you'd have to go through and understand which errors represent valid problems related to user input and which errors represent the system being in a bad and unaccounted for state.

ZenGround0: Some other things that I think meet this criterion: `InvalidRateChangeQueueState`, `LockupInconsistencyDuringRailFinalization`, `LockupExceedsFundsInvariant`

src/Payments.sol
```diff
@@ -1273,7 +1343,9 @@ contract Payments is Initializable, UUPSUpgradeable, OwnableUpgradeable, Reentra
     }
 
     function remainingEpochsForTerminatedRail(Rail storage rail) internal view returns (uint256) {
```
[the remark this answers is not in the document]
ZenGround0: Actually what I would do is add `validateRailNotTerminated` modifier to this function, problem solved.

src/Payments.sol
ZenGround0: I would consider adding `validateRailNotTerminated` modifier to `remainingEpochsForTerminatedRail` and `maxSettlementEpochForTerminatedRail` function calls.

I think its ok to add `railId` param to isRailTerminated function call.

pali101:
> Hey @pali101 please move back to using require syntax on errors. As @Kubuxu noted [here](https://github.com/FilOzone/filecoin-services-payments/pull/166#discussion_r2182725023) this is possible with builtin require in solidity.

Sure, will do.

However, support for `require(condition, CustomError())` was only introduced in Solidity `0.8.26`. Our current `foundry.toml` specifies `0.8.25`, so we’ll need to upgrade the compiler version before we can use this pattern.

Is there any particular reason we were using Solidity `0.8.25` and not newer version?

ZenGround0:
> Is there any particular reason we were using Solidity 0.8.25 and not newer version?

I don't think so, updating sounds like a good idea.  @aarshkshah1992 can correct me if I'm wrong once he comes back in which case we can revert and do a user defined require approach.

src/Payments.sol
pali101: Done it for most of the errors except `RateChangeQueueNotEmpty()` since inside require it evaluates all the conditions and I'm returning tail element of queue so even if queue is empty it tries to peek tail.

PR #167, opened 2025-07-02
Termination language

ZenGround0:
Small README fixup

### DIFF
--- a/README.md
+++ b/README.md
@@ -364,7 +364,7 @@ Payments.modifyRailPayment(
   1. **Rail Active:**  
      While the rail is active, the operator can make one-time payments at any time, provided there is sufficient fixed lockup remaining.
   2. **Rail Termination:**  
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -248,7 +248,7 @@ Retrieves the current state of a payment rail.
 
 #### `terminateRail(uint256 railId)`
 
-Emergency termination of a payment rail, preventing new payments after the lockup period. This should only be used in exceptional cases where the operator contract is malfunctioning and refusing to cancel deals.
+Normal termination of payment that enters the rail into its final lockup period before finalizing completion of the payment flow. This can be called by the operator or a client in good standing.
```
rjan90, line 251: This sentence seemed to be a bit of a mouthful, at least the bolded out parts of it:
>Normal termination of payment that enters the rail **into its final lockup period before finalizing completion of the payment flow.**

But this might also just be that English is not my native language.

ZenGround0: All good docs should be easy for anyone who is ESL so this is good feedback.  I'll make it better

rjan90: No it reads better for me at least 🙏

PR #172, opened 2025-07-10
Spec for auditing

ZenGround0:
Spec to help auditors or other development contributors to onboard into the payments contract implementation

TODO
- [x] Rail settlement
- [x] Other things about rails (validation, mods, OTP) etc
- [x] Termination


src/Payments.sol
```diff
@@ -289,9 +289,9 @@ contract Payments is Initializable, UUPSUpgradeable, OwnableUpgradeable, Reentra
         require(
             !settleFull || isAccountLockupFullySettled(payer),
             isBefore
-                ? "payers's account lockup target was not met as a precondition of the requested operation"
+                ? "payers's full account lockup was not met as a precondition of the requested operation"
                 :
-                "the requested operation would cause the payer's account lockup target to exceed the funds available in the account"
```
ZenGround0, line 294: I'm pretty sure this is incorrect in all cases since we check for this invariant in `settleAccountLockup`

SPEC.md
```diff
@@ -0,0 +1,271 @@
...
+
+Some concepts are a bit tricky and show up throughout the code in subtle ways. Once you understand them it makes things easier.
+
+### Three Core Data Structures
+
+There are three essential data structures in this contract.  The [`Account`](#accounts-and-account-settlement), the [`Rail`](#rails-and-rail-settlement) and the [`OperatorApproval`](#operator-approval). Accounts hold funds of a particular token associated with a public key. They are used for paying and receiving payment. Rails are used to track point to point payments between Accounts. OperatorApprovals allow an operator contract to set up and modify payments between parties under usage constraints.
+
+A public key identity can have multiple Accounts of different token type. Each Account can have multiple operators that it has approved to process payments. Each Account can also have multiple outgoing payment rails. Each rail represents a different payee. There is one operator per rail. One operator can manage many rails and each rail can have a different operator. To consider the general picture it can be helpful to think of a set of operators per account and a set of rails per operator.
```
aarshkshah1992, line 29: "Each Account can also have multiple outgoing and incoming payment rails"*

SPEC.md
```diff
@@ -0,0 +1,271 @@
...
+
+Rail payment flows on terminated rails are locked and known as the streaming lockup. These funds are locked when `modifyRailPayment` increases the rail's payment rate or when `modifyRailLockup` changes the lockup period.  They are released during settlement of the terminated rail.  This is a very essential point to understand the payments contract.  Rate based payments paid out during the `lockupPeriod` for a terminated rail share characteristics of both one time payments and live rail payment streams.  Like one time payments all rails are required to lockup up front the amount needed to cover the lockup period payment.  Like live rail payments the `lockupPeriod` payments are released at the rail's rate through time. Unique to rail payments after termination is that they *must* flow from payer to payee, barring validation interference.  One time payments have no such requirement and live rail payments can always be stopped by terminating the rail.
```
aarshkshah1992, line 43: Would be nice to document in a way that it is easily evident  that for non-terminate rails, payees can never withdraw the `rate * lockup period` funds as that is an escrow guarantee only withdrawable after a rail is terminated.

SPEC.md
```diff
@@ -0,0 +1,271 @@
...
+        uint256 lockupUsage; // Track actual usage for lockup
+        uint256 maxLockupPeriod; // Maximum lockup period the operator can set for rails created on behalf of the client
+    }
+```
+
+An important counterintuitive fact about the approval allowances is that they are not constrained in relation to current usage. Usage can be lower than allowance if an operator has not used all of their existing allowance. Usage can be higher than allowance if a client has manually reduced the operator's allowance. As explained in the README reducing allowance below usage on any of the allowance resources (rate, lockup, period) will not impact existing rails. Allowance invariants are checked at the point in time of rail modification not continuously enforced. Furthermore reductions in usage always go through even if the current allowance is below the new usage. For example if a rail has an allowance of 20 locked tokens and uses all of them to lock up 20 tokens, and then the client brings allowance for the operator down to 1 locked token the operator can still modify the rail usage down to 15 locked tokens even though it exceeds the operator's current allowance.
```
aarshkshah1992, line 122: However, new rails created after a "a client has manually reduced the operator's allowance" will have to respect the new allowance limits

PR #178, opened 2025-07-16
Zero segment bug

ZenGround0:
The queue is processed in a fun way such that it prematurely pops off queue segments.  As a consequence the rail can be tricked into paying for the remainder of a partially completed segment at the next segment's rate.  While at first I thought this was confined only to zero segments it is actually a problem for both the zero and nonzero case.

The fix I've added only removes segments from the queue if they have been completely processed.


aarshkshah1992 (approved):
nice

lordshashank (approved):
Nice find!!!

PR #182, opened 2025-07-18
chore(doc): minor README cleanups and fixups

rvagg:
* "SP" -> "service provider"
* document EIP-2612 options
* fix method signature discrepancies
* \s+$
* misc cleanup
* update TOC
* add license in README

(Initially I just wanted to deal with "SP" but then I couldn't stop!)


README.md
```diff
@@ -666,5 +741,10 @@ Payments.modifyRailPayment(
 ```
 
 #### Best Practice
+
 - Unless absolutely required, clients should refrain from cutting operator allowances for ongoing transactions.
 - In the event that the rate stream cannot be initiated, operators should be prepared for this possibility and utilize one-time payments as a backup.
+
+## License
+
+Licensed under [MIT](https://github.com/filecoin-project/lotus/blob/master/LICENSE-MIT)
```
rjan90, line 750:
```suggestion
Licensed under [MIT](https://github.com/FilOzone/filecoin-services-payments/blob/main/LICENSE)
```

rjan90 (approved):
Looks good to me, one small nit on the link

README.md
rjan90: Another question is if we should add a dual-licence similar to the PDP-contract: https://github.com/FilOzone/pdp?tab=License-1-ov-file#readme

rjan90: Okay, you beat me to it: https://github.com/FilOzone/filecoin-services-payments/pull/183

PR #183, opened 2025-07-18
doc(license): dual MIT / Apache 2.0

rvagg:
This should probably have been done in the first place when UNLICENSED was replaced. Dual licensing has typically been our practice, but having Apache 2.0 in here is pretty important because it's often preferred by larger orgs that have concerns related to patents.


jennijuju (approved):
oops thanks!

PR #191, opened 2025-07-26
Add automated interface generation for Payments contract

### LINKED ISSUE #122: Interface for payments contract
It would be nice to have `IPayments.sol` for payments contract which can be used by devs for inheritance, etc.

pali101:
## Summary

This PR introduces an auto-generated `IPayments` interface for the Payments contract, along with tooling and CI checks to ensure it remains consistent with the contract ABI.

## Changes

- Generated `src/interfaces/IPayments.sol` using [abi-to-sol](https://github.com/gnidan/abi-to-sol) for external use and inheritance
- Script: `tools/generate-interface.sh` builds the contract, extracts its ABI, and generates a clean IPayments interface
- Refactor: Moved `IValidator` to `src/interfaces/IValidator.sol`
- `Makefile`: Added `make interface` target
- CI: GitHub Action to check interface drift on push/PR to main
- Docs: Updated `tools/README.md` with usage details for the script and make interface

Closes #122


tools/generate-interface.sh
```diff
@@ -0,0 +1,48 @@
...
+
+# Generate the interface using abi-to-sol
+npx --prefix "$TEMP_DIR" abi-to-sol IPayments \
+  --license MIT \
+  --solidity-version "^0.8.27" \
+  < "$TEMP_DIR/Payments.abi.json" > "$OUTPUT_PATH"
+
+# Remove ABI comments from the generated file
+sed -i '/^\/\/ THIS FILE WAS AUTOGENERATED FROM THE FOLLOWING ABI JSON:/,/^\*\//d' "$OUTPUT_PATH"
+sed -i 's/SEE SOURCE BELOW\. *!!/!!/' "$OUTPUT_PATH"
```
rvagg, line 41: these are not going to be portable, on macos `-i` does a different thing, I think you'll need `-i ''` if you detect macos; it would be nice if these were portable on both linux & darwin so anyone can run them

tools/generate-interface.sh
```diff
@@ -0,0 +1,48 @@
...
+for cmd in forge jq npm; do
+  if ! command -v $cmd &> /dev/null; then
+    echo "Error: '$cmd' is not installed or not in PATH." >&2
+    exit 1
+  fi
+done
+
+# Build the contracts
+echo "Building contracts..."
+forge build --force > /dev/null 2>&1
```
rvagg, line 16: can we capture errors? maybe something like:

```suggestion
forge build --force 2>&1 | grep -Ei "^error" || true
```

(though I'm not too familiar with forge to know what errors consistently look like so maybe check?)

rvagg: or .. just let all the output past, is it really too much?

tools/generate-interface.sh
```diff
@@ -0,0 +1,48 @@
...
+
+# Create a temporary directory and install abi-to-sol locally
+TEMP_DIR=$(mktemp -d)
+cd "$TEMP_DIR"
```
rvagg, line 20: you could get rid of this TEMP_DIR thing entirely and just pipe, combining with `npx` below you could end up with something like:

```
jq '.abi' out/Payments.sol/Payments.json | npx abi-to-sol@0.8.0 IPayments --license MIT --solidity-version "^0.8.27" > "$OUTPUT_PATH"
```

tools/generate-interface.sh
```diff
@@ -0,0 +1,48 @@
+#! /bin/bash
+# generate-interface.sh generates the interface for the Payments contract
+
+set -euo pipefail
+
+# Check for required tools
+for cmd in forge jq npm; do
```
rvagg, line 7:
```suggestion
for cmd in forge jq npx; do
```

Makefile
```diff
@@ -40,6 +40,10 @@ chmod-transfer:
 chmod-get-owner:
 	chmod +x ./tools/get-owner.sh
 
+.PHONY: chmod-interface
+chmod-interface:
+	chmod +x ./tools/generate-interface.sh
+
```
rvagg, line 46:
```suggestion
```

rvagg: if you're really worried about this, just prefix with `bash` below

tools/generate-interface.sh
pali101: Updated the script to pipe the ABI directly into npx, completely removed the temp dir and init logic. Learned something new here, thanks @rvagg!

tools/generate-interface.sh
pali101: Done

tools/generate-interface.sh
pali101: Thanks! Added conditional logic to handle both `linux` and `mac`.

.github/workflows/check-interface.yml
```diff
@@ -0,0 +1,37 @@
...
+          sudo apt-get install -y nodejs
+
+      - name: Verify interface
+        run: |
+          make interface
+          git diff --exit-code src/interfaces/IPayments.sol || \
+            (echo "Interface out of sync. Run 'make interface' locally and commit the changes."; exit 1)
+
+      - name: Clean up temporary files
+        run: rm -rf temp
```
rvagg, line 37:
```suggestion
```

tools/generate-interface.sh
```diff
@@ -0,0 +1,38 @@
...
+# Build the contracts
+echo "Building contracts..."
+forge build --force
+
+# Create dir if it doesn't exist
+mkdir -p "$(dirname "$OUTPUT_PATH")"
+
+# Extract ABI using jq
+echo "Generating interface at $OUTPUT_PATH..."
+jq '.abi' out/Payments.sol/Payments.json | npx abi-to-sol@0.8.0 IPayments --license MIT --solidity-version "^0.8.27" > "$OUTPUT_PATH"
```
rvagg, line 25: optional, but you could pull these two versions up to variables at the top of the file so we can update it more easily, but it's not a big deal, we'll probably only ever change this when stuff doesn't work the way we expect -- unless we have the solidity version hard-wired somewhere else too?

aarshkshah1992:
@pali101 Let's keep this PR open till we integrate it with Pandora and make sure the interface/compilation/tests work in Pandora as well.

BigLep:
@rjan90 @pali101 : what are the next steps.  Is this still worth landing?

PR #192, opened 2025-07-28
View: Truncate instead of Copy

wjmelements:
Reviewer @aarshkshah1992
Solidity memory arrays cannot increase in size once allocated, but it safe to truncate them.
Because memory gas is quadratic, the cost of this copy grows quickly with the size of the array.
By truncating the array we can avoid the copy step.
#### Changes
* replace array copy with array truncation
### Test plan
* verified that without truncation at least one test fails

### DIFF
--- a/src/Payments.sol
+++ b/src/Payments.sol
@@ -1581,7 +1581,7 @@ contract Payments is Initializable, UUPSUpgradeable, OwnableUpgradeable, Reentra
             // Skip non-existent rails
             if (rail.from == address(0)) continue;
 
-            // Add rail to our temporary array
+            // Add rail info to results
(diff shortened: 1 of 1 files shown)


wjmelements:
> memory gas is quadratic

I'm not sure that this is the case for Filecoin but I still think this is a good change.

For Ethereum, here is the gas savings by number of results:
| Results    | Savings |
| -------- | ------- |
| 0 | -22 |
| 300  |  310600  | 
| 600 | 654079 |
| 900 | 1031309 |
| 1200 | 1442287 |
| 1500 | 1887016 |
| 1800 | 2365495 |
| 2100 | 2877724 |

Issue #127, opened 2025-06-17
Investigate Gas Costs and BaseFee Scenario Viability


rjan90:
## Done Criteria:

**Gas Cost Analysis:** 
- Collect current gas cost metrics for key Payments Contract operations:
  - Account management (deposit, withdraw, withdrawTo)
  - Operator management (setOperatorApproval)
  - Rail management (createRail, modifyRailLockup, modifyRailPayment, terminateRail)
  - Settlement operations (settleRail, settleTerminatedRailWithoutArbitration)
  - Rail query operations (getRail, getRailsForPayerAndToken, getRailsForPayeeAndToken)

**BaseFee Scenario Testing:** 
- Simulate Payments Contract operations under different BaseFee scenarios (e.g., low/medium/high network congestion)
- Test gas consumption for:
  - Single rail vs. multiple rail scenarios
  - Different token types and amounts
  - Various lockup periods and payment rates
  - Settlement operations with and without arbitration
- Identify thresholds where gas costs become prohibitive for users, service providers, or operators

## Why Important:
Ensures Payments Contract operations remain viable under varying network conditions, and prevents unexpected cost spikes from disrupting client/SP payment workflows. Helps inform fee structure and economic model design

Chaitu-Tatipamula:
i can take this up. can you assign it tome?

rvagg:
> Forge gas report

How does this work? Is it looking at opcodes and making assumptions about gas cost, or is it looking at recent messages on chain? Unfortunately the Filecoin gas model is quite different to Ethereum so I'm not sure how well that will translate to Filecoin gas costs.

Chaitu-Tatipamula:
- Forge executes the actual contract code and measures gas usage in computational units
- Filecoin's FVM uses the same EVM interpreter and opcodes as Ethereum so the gas usage should be identical between networks

well the calculation is actually different
Filecoin uses: `(Gas usage × Base fee) + (GasLimit × GasPremium) + (OverEstimationBurn × BaseFee)`

so the Gas usage (what was measured) should be same on both Ethereum and Filecoin

Chaitu-Tatipamula:
> > Filecoin's FVM uses the same EVM interpreter and opcodes as Ethereum
> 
[...]

PR #193, opened 2025-07-28
Gas Snapshot

wjmelements:
Reviewer @aarshkshah1992

Gas roughly measures the burden of operations on the network, so we want to reduce it.

Although filecoin gas won't be the same as Ethereum, measuring the ethereum gas will provide a way to know to what extent a change reduces gas.

#### Changes
* add .gas-snapshot
* add workflow to check the .gas-snapshot


.gas-snapshot
```diff
@@ -94,8 +94,9 @@ OperatorApprovalUsageLeakTest:testOperatorLockupUsageLeakOnRailFinalization() (g
 PayeeFaultArbitrationBugTest:testLockupReturnedWithFault() (gas: 910072)
 PayeeFaultArbitrationBugTest:testLockupReturnedWithFaultReducedDuration() (gas: 2059695)
 PayeeFaultArbitrationBugTest:testLockupReturnedWithFaultTermination() (gas: 888076)
-PayeeRailsTest:testEmptyResult() (gas: 57771)
-PayeeRailsTest:testGetRailsForPayeeAndToken() (gas: 149980)
+PayeeRailsTest:testEmptyResult() (gas: 57793)
+PayeeRailsTest:testGetManyRails() (gas: 130853979)
```
wjmelements, line 98: looks like creating new rails is especially expensive

rvagg:
Some previous gas discussions for context: https://github.com/FilOzone/filecoin-services-payments/issues/127 - leading to `docs/payments-gas-benchmarks.md` in https://github.com/FilOzone/filecoin-services-payments/pull/175

wjmelements:
> Some previous gas discussions for context: #127 - leading to `docs/payments-gas-benchmarks.md` in #175

Is there an easy way to update that?

aarshkshah1992:
@wjmelements Wdym ? You can just open a PR to update those numbers as the doc lives in this repo (pretty sure I'm missing something here).

wjmelements:
> @wjmelements 
> 
> 
> 
> """
> 
> _Gas roughly measures the burden of operations on the network, so we want to reduce it._
> 
> """
> 
> For my own understanding:
> 
> 
> 
> Dosen't it also help increase network revenue for token holders if the gas token is the same as chain token ?
> 
> All for decreasing gas as it improves UX but I've never been able to reconcile improving UX for end-users with decreasing revenue for token holders in my mental model.

If the payments are more efficient, the network can do more of them, and we can supply them more sustainably and competitively. This is the supply side.

If payments are too expensive, existing and potential users will  choose cheaper alternatives. This is the demand side.

wjmelements:
Closing in favor of #207

PR #202, opened 2025-08-05
Fix Large Contract

wjmelements:
Reviewer @rvagg
[explanation](https://github.com/FilOzone/filecoin-services-payments/blob/main/docs/migration-to-foundry-v1.md)
#### Changes
* optimizer on


wjmelements:
Payments.sol:Payments size 
* before: 33399
* after: 16863

PR #203, opened 2025-08-05
 Comprehensive Builders manual for payments contract

aarshkshah1992:
Do a deep dive surgery of the README to create a comprehensive builders manual for payments contract.

### DIFF
--- a/.github/scripts/check-contract-size.sh
+++ b/.github/scripts/check-contract-size.sh
@@ -36,6 +36,13 @@ if ! jq empty contract_sizes.json 2>/dev/null; then
     exit 1
 fi
 
+if jq -e '. == {}' contract_sizes.json >/dev/null; then
+    echo "forge did not find any contracts. forge build:"
+    # This usually means build failure
+    forge build
+    exit 1
+fi
+
 json=$(cat contract_sizes.json)
 
 # Filter JSON: keep only contracts/libraries from src/
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# FWS Payments Contract
+# Filecoin Pay
 
 The Filecoin Pay Payments contract enables ERC20 token payment flows through "rails" - automated payment channels between payers and recipients. The contract supports continuous rate based payments, one-time transfers, and payment validation during settlement.
 
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # FWS Payments Contract
 
-The FWS Payments contract enables ERC20 token payment flows through "rails" - automated payment channels between payers and recipients. The contract supports continuous rate based payments, one-time transfers, and payment validation during settlement.
+The Filecoin Pay Payments contract enables ERC20 token payment flows through "rails" - automated payment channels between payers and recipients. The contract supports continuous rate based payments, one-time transfers, and payment validation during settlement.
 
 - [Deployment Info](#deployment-info)
 - [Key Concepts](#key-concepts)
--- a/README.md
+++ b/README.md
@@ -63,9 +63,11 @@ Tracks the funds, lockup, obligations, etc. associated with a single “owner”
 
 ### Rail
 
-A rail along which payments flow from a payer to a payee. Rails track lockup, maximum payment rates, and obligations between a payer and a servive provider. Payer ↔ Payee pairs can have multiple payment rails between them but they can also reuse the same rail across multiple deals. Importantly, rails:
+A rail along which payments flow from a payer to a payee. Rails track lockup, maximum payment rates, and obligations between a payer and a payee. Payer ↔ Payee pairs can have multiple payment rails between them but they can also reuse the same rail across multiple deals. Importantly, rails:
 - Specify the maximum rate at which the payer will pay the payee, the actual amount paid for any given period is subject to validation by the **validator** described below.
-- Define the required funding commitment for future payments through the **lockup period**. While the contract cannot force a payer to deposit funds from their external wallet, it strictly enforces lockups on all funds held within their contract account. It prevents payers from withdrawing locked funds and blocks operator actions that would increase a payer's lockup obligation beyond their available balance. This system provides a reliable way for payees to verify a payer's funding commitment for the agreed-upon period.
+- Define a lockup period. The lockup period of a rail is the time period over which the payer is required to maintain locked funds to fully cover the current outgoing payment rate from the rail if the payer stops adding funds to the account. This provides a reliable way for payees to verify that a payer is guaranteed to pay up to a certain point in the future. When a rail's payer account drops to only cover the lockup period this is a signal to the payee that the payer is at risk of defaulting. The lockup period gives the payee time to settle and gracefully close down the rail without missing payment.
+- Strictly enforce lockups. While the contract cannot force a payer to deposit funds from their external wallet, it strictly enforces lockups on all funds held within their contract account. It prevents payers from withdrawing locked funds and blocks operator actions that would increase a payer's lockup obligation beyond their available balance. This system provides an easy way for payees to verify a payer's funding commitment for the rail.
+
 
 ### Validator
 
@@ -79,7 +81,7 @@ When a validator is assigned to a rail, it gains the ability to:
 
 ### Operator
 
-An operator is a smart contract—typically the main contract for a given service—that manages payment rails on behalf of payers. A payer must explicitly approve an operator and grant it specific allowances, which act as a budget for how much the operator can spend or lock up on the payer's behalf.
+An operator is a smart contract (typically the main contract for a given service) that manages payment rails on behalf of payers. It is also sometimes referred to as the "service contract". A payer must explicitly approve an operator and grant it specific allowances, which act as a budget for how much the operator can spend or lock up on the payer's behalf.
 
 The operator role is powerful, so the operator contract must be trusted by both the payer and the payee. The payer trusts it not to abuse its spending allowances, and the payee trusts it to correctly configure and manage the payment rail.
 
@@ -88,12 +90,12 @@ An approved operator can perform the following actions:
 -   **Create Rails (`createRail`):** Establish a new payment rail from a payer to a payee, specifying the token, payee, and an optional validator.
 -   **Modify Rail Terms (`modifyRailLockup`, `modifyRailPayment`):** Adjust the payment rate, lockup period, and fixed lockup amount for any rail it manages. Any increase in the payer's financial commitment is checked against the operator's allowances.
 -   **Execute One-Time Payments (`modifyRailPayment`):** Execute one-time payments from the rail's fixed lockup.
--   **Settle Rails (`settleRail`):** Trigger payment settlement for a rail to process due payments. As a rail participant, the operator can initiate settlement at any time.
+-   **Settle Rails (`settleRail`):** Trigger payment settlement for a rail to process due payments within the existing terms of the rail. As a rail participant, the operator can initiate settlement at any time. The operator cannot, however, arbitrarily settle a rail for a higher-than-expected amount or higher than expected duration.
 -   **Terminate Rails (`terminateRail`):** End a payment rail. Unlike payers, an operator can terminate a rail even if the payer's account is not fully funded.
 
 ### Per-Rail Lockup: The Guarantee Mechanism
 
-Each payment rail requires the payer to lock funds to guarantee future payments. This lockup is composed of two distinct components:
+Each payment rail can be configured to require the payer to lock funds to guarantee future payments. This lockup is composed of two distinct components:
 
 -   **Streaming Lockup (`paymentRate × lockupPeriod`):** A calculated guarantee for rate based payments for a pre-agreed lockup period.
 -   **Fixed Lockup (`lockupFixed`):** A specific amount set aside for one-time payments.
@@ -106,7 +108,7 @@ It is critical to understand that the streaming lockup is **not** a pre-paid acc
 
 **1. During Normal Operation (Before Termination)**
 
-While a rail is active, the streaming lockup acts as a **guarantee of solvency**, not a direct source of payment.
+While a rail is active, the streaming lockup acts as a **guarantee of solvency for a pre-agreed number of epochs**, not as a direct source of payment.
 
 -   **Payments from General Funds:** When `settleRail` is called on an active rail, payments are drawn from the payer's general `funds`.
 -   **Lockup as a Floor:** The lockup simply acts as a minimum balance. The contract prevents the payer from withdrawing funds below this floor.
@@ -161,9 +163,9 @@ The operator reduces the lockup period to 5 epochs.
 
 This lockup mechanism places clear responsibilities on the payee to manage risk:
 
--   **Settle Regularly:** You must settle rails frequently. A rail's `lockupPeriod` is a measure of the risk you are willing to take. If you wait longer than the `lockupPeriod` to settle, you allow a payer to build up a payment obligation that may not be fully covered by the lockup guarantee if they become insolvent.
+-   **Settle Regularly:** Depending on the solvency guarantees put in place by the operator contract's lockup requirements, you must settle rails frequently. A rail's `lockupPeriod` is a measure of the risk you are willing to take. If you wait longer than the `lockupPeriod` to settle, you allow a payer to build up a payment obligation that may not be fully covered by the lockup guarantee if they become insolvent.
 -   **Monitor Payer Solvency:** Use the `getAccountInfoIfSettled` function to check if a payer is funded. If their `fundedUntilEpoch` is approaching the current epoch, they are at risk.
--   **Terminate Proactively:** If a payer becomes insolvent or unresponsive, terminate the rail immediately. This is the **only way** to activate the safety hatch and ensure you can claim payment from the funds guaranteed by the streaming lockup.
+-   **Terminate Proactively:** If a payer becomes insolvent or unresponsive, request the operator to terminate the rail immediately. This is the **only way** to activate the safety hatch and ensure you can claim payment from the funds guaranteed by the streaming lockup.
 
 ## Core Functions
 
@@ -175,31 +177,37 @@ Functions for managing user accounts, including depositing and withdrawing funds
 
 Deposits tokens into a specified account. This is the standard method for funding an account if not using permits. It intelligently handles fee-on-transfer tokens by calculating the actual amount received by the contract.
 
-- **When to use:** Use this for direct transfers from a wallet or another contract that has already approved the Payments contract to spend tokens.
-- **Native Token (FIL):** To deposit the native network token, use `address(0)` for the `token` parameter and send the corresponding amount in the transaction's `value`.
-- **Parameters**:
-  - `token`: ERC20 token contract address (`address(0)` for FIL).
-  - `to`: The account address to credit with the deposit.
-  - `amount`: The amount of tokens to transfer.
-- **Requirements**:
-  - For ERC20s, the direct caller (`msg.sender`) must have approved the Payments contract to transfer at least `amount` of the specified `token`.
-  - For the native token, `msg.value` must equal `amount`.
+**When to use:** Use this for direct transfers from a wallet or another contract that has already approved the Payments contract to spend tokens.
+
+**Native Token (FIL):** To deposit the native network token, use `address(0)` for the `token` parameter and send the corresponding amount in the transaction's `value`.
+
+**Parameters**:
+- `token`: ERC20 token contract address (`address(0)` for FIL).
+- `to`: The account address to credit with the deposit.
+- `amount`: The amount of tokens to transfer.
+
+**Requirements**:
+- For ERC20s, the direct caller (`msg.sender`) must have approved the Payments contract to transfer at least `amount` of the specified `token`.
+- For the native token, `msg.value` must equal `amount`.
 
 #### `depositWithPermit(address token, address to, uint256 amount, uint256 deadline, uint8 v, bytes32 r, bytes32 s)`
 
 Deposits tokens using an EIP-2612 permit, allowing for gasless token approval.
 
-- **When to use:** Ideal for user-facing applications where the user can sign a permit off-chain. This combines approval and deposit into a single on-chain transaction, saving gas and improving user experience.
-- **Note:** This function is for ERC20 tokens only and does not support the native token.
-- **Parameters**:
-  - `token`: ERC20 token contract address supporting EIP-2612 permits.
-  - `to`: The account address to credit (must be the signer of the permit).
-  - `amount`: Token amount to deposit.
-  - `deadline`: Permit expiration timestamp.
-  - `v`, `r`, `s`: Signature components for the EIP-2612 permit.
-- **Requirements**:
-  - Token must support EIP-2612.
-  - `to` must be `msg.sender` (the one submitting the transaction).
+**When to use:** Ideal for user-facing applications where the user can sign a permit off-chain. This combines approval and deposit into a single on-chain transaction, saving gas and improving user experience.
+
+**Note:** This function is for ERC20 tokens only and does not support the native token.
+
+**Parameters**:
+- `token`: ERC20 token contract address supporting EIP-2612 permits.
+- `to`: The account address to credit (must be the signer of the permit).
+- `amount`: Token amount to deposit.
+- `deadline`: Permit expiration timestamp.
+- `v`, `r`, `s`: Signature components for the EIP-2612 permit.
+
+**Requirements**:
+- Token must support EIP-2612.
+- `to` must be `msg.sender` (the one submitting the transaction).
 
 #### `depositWithPermitAndApproveOperator(...)`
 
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -48,298 +48,400 @@ The FWS Payments contract enables ERC20 token payment flows through "rails" - au
...
+  - Decide the final financial outcome (the total payout) of a rail that has been successfully terminated.
 - **Operator**: An authorized third party who can manage rails on behalf of clients
 
 ### Account
 
 Tracks the funds, lockup, obligations, etc. associated with a single “owner” (where the owner is a smart contract or a wallet). Accounts can be both *clients* and *service providers* but we’ll often talk about them as if they were separate types.
 
 - **Client —** An account that *pays* a service provider (also referred to as the *payer*)
-- **Service Provider** — An account managed the provider of some service which receives payment from a client (also referred to as the *payee*).
+- **Service Provider** — An account managed by the provider of some service which receives payment from a client (also referred to as the *payee*).
```
ZenGround0, line 62: nit: To be honest the notion of client and provider feels secondary to the Payer Payee concept.  Services and clients are an application of payments and mixing that into the terminology here has always felt confusing.  If you agree and are feeling inspired you could swap the primary and secondary names around.

README.md
```diff
@@ -48,298 +48,400 @@ The FWS Payments contract enables ERC20 token payment flows through "rails" - au
...
 
-An validator is an (optional) smart contract that can validate / arbitrate payments associated with a single rail. For example, a payment rail used for PDP will specify the PDP service as its validator. A validator can:
+A validator is an optional contract that acts as a trusted arbitrator for a rail. Its primary role is to validate payments during settlement, but it also plays a crucial part in the rail's lifecycle, especially during termination.
 
-- Prevent settlement of a payment rail entirely.
-- Refuse to settle a payment rail past some epoch.
-- Reduce the amount paid out by a rail for a period of time (e.g., to account for actual services rendered, penalties, etc.).
+When a validator is assigned to a rail, it gains the ability to:
+
+-   **Mediate Payments:** During settlement, a validator can prevent a payment, refuse to settle past a certain epoch, or reduce the payout amount to account for actual services rendered, penalties, etc.
```
ZenGround0, line 76: nit: I've found the deleted bullet list useful.  Maybe restructure them to sub-bullets here.

README.md
```diff
@@ -48,298 +48,400 @@ The FWS Payments contract enables ERC20 token payment flows through "rails" - au
...
+-   **Fixed Lockup (`lockupFixed`):** A specific amount set aside for one-time payments.
 
-```
-Total Lockup Required = (paymentRate × lockupPeriod) + lockupFixed
-```
+The total lockup for a client's account is the sum of these requirements across *all* their active rails. This total is reserved from their deposited funds and cannot be withdrawn.
+
+#### The Crucial Role of Streaming Lockup: A Safety Hatch, Not a Pre-payment
+
+It is critical to understand that the streaming lockup is **not** a pre-paid account that is drawn from during normal operation. Instead, it functions as a **safety hatch** that can only be fully utilized *after* a rail is terminated.
```
ZenGround0, line 105: 👍

README.md
```diff
@@ -48,298 +48,400 @@ The FWS Payments contract enables ERC20 token payment flows through "rails" - au
...
 - Specify the maximum rate at which the client will pay the service provider, the actual amount paid for any given period is subject to validation by the **validator** described below.
-- Specify the period in advance the client is required to lock funds (the **lockup period**). There’s no way to force clients to lock funds in advance, but we can prevent them from *withdrawing* them and make it easy for service providers to tell if their clients haven’t met their lockup minimums, giving them time to settle their accounts.
+- Define the required funding commitment for future payments through the **lockup period**. While the contract cannot force a client to deposit funds from their external wallet, it strictly enforces lockups on all funds held within their contract account. It prevents clients from withdrawing locked funds and blocks operator actions that would increase a client's lockup obligation beyond their available balance. This system provides a reliable way for service providers to verify a client's funding commitment for the agreed-upon period.
```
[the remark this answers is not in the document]
ZenGround0: After reading the section on streaming lockup below perhaps the first bullet is going into more detail than necessary

README.md
```diff
@@ -48,298 +48,400 @@ The FWS Payments contract enables ERC20 token payment flows through "rails" - au
...
-  - For active rails: changes restricted if client's account isn't fully funded
-  - Operator must have sufficient allowances
+-   **When to use:** An operator calls this to adjust the client's funding guarantee. This is used to set an initial `lockupFixed` for an onboarding fee, increase the `lockupPeriod` for a longer-term commitment, or decrease lockups when a deal's terms change.
+
+**Lockup Calculation and State Changes**
+
+This function recalculates the rail's total lockup requirement based on the new `period` and `lockupFixed` values. The change in the rail's individual lockup is then applied to the client's total account lockup (`Account.lockupCurrent`).
+
+-   **State Impact:** It modifies both the `Rail` struct (updating `lockupPeriod` and `lockupFixed`) and the client's `Account` struct (updating `lockupCurrent`).
+-   **Safety Checks:** The entire operation is wrapped in a modifier that settles the client's account lockup before and after the change. This ensures that any increase in lockup is only permitted if the client has sufficient available funds to cover it, maintaining the contract's solvency invariants.
```
ZenGround0, line 376: I would probably remove mention of the modifier mechanism.  It's more implementation detail than the rest of the readme goes into and covered in the spec.

README.md
```diff
@@ -48,298 +48,400 @@ The FWS Payments contract enables ERC20 token payment flows through "rails" - au
...
-  - `newRate`: New per-epoch payment rate
-  - `oneTimePayment`: Optional immediate payment amount
-- **Requirements**:
-  - Caller must be the rail operator
-  - For terminated rails: cannot increase rate
-  - For active rails: rate changes restricted if client's account isn't fully funded
-  - One-time payment must not exceed fixed lockup
+-   **When to use:** This is the primary function for starting a payment stream (by setting an initial `newRate`), adjusting it, or making ad-hoc payments.
+
+**Rate Change Behavior**
```
ZenGround0, line 395: Probably there should be a **One Time Payment Behavior** section?

README.md
```diff
@@ -437,36 +539,42 @@ A client can reduce the operator's `maxLockupPeriod` or `lockupAllowance` after
...
-  - `finalSettledEpoch`: Epoch to which settlement was completed
-  - `note`: Additional settlement information
-- **Requirements**:
-  - Client must have sufficient funds to cover the payment
-  - Client's account must be fully funded _or_ the rail must be terminated
-  - Cannot settle future epochs
+The behavior of `settleRail` critically depends on whether the rail is active or terminated:
+
+-   **For Active Rails:** Settlement can only proceed up to the epoch the client's account was last known to be fully funded (`lockupLastSettledAt`). This is a key safety feature: if a client becomes insolvent, settlement of an active rail halts, preventing it from running a deficit.
+-   **For Terminated Rails:** Settlement can proceed up to the rail's final `endEpoch`, drawing directly from the locked funds that act as a safety guarantee.
```
ZenGround0, line 551: nit: "drawing directly from the streaming lockup"

ZenGround0 (approved):
I've left some discussion points and changes that I think would improve this but nothing blocking.

README.md
```diff
@@ -48,298 +48,400 @@ The FWS Payments contract enables ERC20 token payment flows through "rails" - au
...
 
-Each payment rail requires the user to manage two lockup components:
+### Per-Rail Lockup: The Guarantee Mechanism
 
-- **Payment Stream lockup:** `paymentRate × lockupPeriod` (covers future payment streams)
-- **Fixed lockup:** `lockupFixed` (covers one-time payments)
+Each payment rail requires the client to lock funds to guarantee future payments. This lockup is composed of two distinct components:
 
-The contract always enforces that the sum of these two values is locked for each rail:
+-   **Streaming Lockup (`paymentRate × lockupPeriod`):** A calculated guarantee for rate based payments for a pre-agreed lockup period.
```
rvagg, line 98: I'm not sold on "Streaming Lockup" although I don't have an alternative, it would be good to check this with people less deeply involved in this to see if they have ideas. "Payment Stream Lockup" felt a bit closer, "Streaming" alone has too many connotations unrelated to payments that can be brought to bear mentally when parsing this I worry.

README.md
```diff
@@ -48,298 +48,400 @@ The FWS Payments contract enables ERC20 token payment flows through "rails" - au
...
+#### The Crucial Role of Streaming Lockup: A Safety Hatch, Not a Pre-payment
+
+It is critical to understand that the streaming lockup is **not** a pre-paid account that is drawn from during normal operation. Instead, it functions as a **safety hatch** that can only be fully utilized *after* a rail is terminated.
+
+**1. During Normal Operation (Before Termination)**
 
-- You must ensure both components are sufficiently funded.
-- One-time payments are deducted from `lockupFixed`.
-- Streaming payments are covered by the streaming lockup.
+While a rail is active, the streaming lockup acts as a **guarantee of solvency**, not a direct source of payment.
```
rvagg, line 109: I do wonder though how much this matters, this is basically saying that the lockup isn't a "source", it's not a wallet or a pool, it's just a **rule**. Does the mental framing of that matter so much? Because in your illustrations below you show tokens coming out of `funds` but then the lockup also being reduced by the same amount, so it's _6 on one hand, half a dozen on the other_. What you're trying to do with this sentence is make sure the reader's mental model sees it in a particular way, I'm just wondering if this matters at all.

README.md
```diff
@@ -48,298 +48,400 @@ The FWS Payments contract enables ERC20 token payment flows through "rails" - au
...
 
-#### Best Practices
+This lockup mechanism places clear responsibilities on the service provider to manage risk:
 
-- **Plan ahead:** Estimate both your ongoing (streaming) and one-time payment needs and fund both buckets accordingly.
-- **Monitor balances:** Before making a one-time payment or increasing the streaming rate / period, check that your account is sufficiently funded.
-- **Use cases:** Use streaming lockup for predictable, recurring payments. Use fixed lockup for immediate payments.
-
-This model gives flexibility: this can guarantee ongoing service payments while also handling special, immediate payments as needed—all with clear, enforced separation and safety for both parties.
+-   **Settle Regularly:** You must settle rails frequently. A rail's `lockupPeriod` is a measure of the risk you are willing to take. If you wait longer than the `lockupPeriod` to settle, you allow a client to build up a payment obligation that may not be fully covered by the lockup guarantee if they become insolvent.
```
rvagg, line 164: "depending on the solvency guarantees put in place by the operator contract's lockup requirements"?

README.md
```diff
@@ -48,298 +48,400 @@ The FWS Payments contract enables ERC20 token payment flows through "rails" - au
...
-Deposits tokens into a specified account.
+Deposits tokens into a specified account. This is the standard method for funding an account if not using permits. It intelligently handles fee-on-transfer tokens by calculating the actual amount received by the contract.
 
+- **When to use:** Use this for direct transfers from a wallet or another contract that has already approved the Payments contract to spend tokens.
+- **Native Token (FIL):** To deposit the native network token, use `address(0)` for the `token` parameter and send the corresponding amount in the transaction's `value`.
 - **Parameters**:
-  - `token`: ERC20 token contract address
-  - `to`: Recipient account address
-  - `amount`: Token amount to deposit
+  - `token`: ERC20 token contract address.
```
rvagg, line 181:
```suggestion
  - `token`: ERC20 token contract address (`address(0)` for FIL).
```

maybe we should do this inline, I missed it above on the first read, can be repeated below for each call taking a `token`

README.md
```diff
@@ -48,298 +48,400 @@ The FWS Payments contract enables ERC20 token payment flows through "rails" - au
...
+  - For ERC20s, the direct caller (`msg.sender`) must have approved the Payments contract to transfer at least `amount` of the specified `token`.
+  - For the native token, `msg.value` must equal `amount`.
 
 #### `depositWithPermit(address token, address to, uint256 amount, uint256 deadline, uint8 v, bytes32 r, bytes32 s)`
 
-Deposits tokens using EIP-2612 permit.
+Deposits tokens using an EIP-2612 permit, allowing for gasless token approval.
 
+- **When to use:** Ideal for user-facing applications where the user can sign a permit off-chain. This combines approval and deposit into a single on-chain transaction, saving gas and improving user experience.
+- **Note:** This function is for ERC20 tokens only and does not support the native token.
```
rvagg, line 193: > does not support the native token
oh, this is good to know! I would have just assumed it would be smart enough to figure that out; should we just add it there since it should be a pretty easy test?

README.md
```diff
@@ -48,298 +48,400 @@ The FWS Payments contract enables ERC20 token payment flows through "rails" - au
...
-  - `amount`: Token amount to deposit
-  - `deadline`: Permit expiration timestamp
-  - `v`, `r`, `s`: Signature components for EIP-2612 permit signature
-  - `operator`: Address to grant permissions to
-  - `rateAllowance`: Maximum payment rate the operator can set across all rails
-  - `lockupAllowance`: Maximum funds the operator can lock for future payments
-  - `maxLockupPeriod`: Maximum allowed lockup period in epochs
+- **When to use:** This is the most efficient way for a new user to get started. It funds their account and authorizes a service contract (operator) in a single step.
+- **Note:** This function is for ERC20 tokens only.
+- *(Parameters are as documented in the contract interface)*
```
rvagg, line 213: disagree, I'd really like these to be in the README like they were; I buy not putting them in the heading (tho I still would), but at least list them

README.md
```diff
@@ -48,298 +48,400 @@ The FWS Payments contract enables ERC20 token payment flows through "rails" - au
...
+  - Amount must not exceed the caller's unlocked funds.
 
 #### `getAccountInfoIfSettled(address token, address owner)`
 
-Displays information about account's current solvency assuming settlement of all active rails
+This is a key read-only function that provides a real-time snapshot of an account's financial health. It works by performing an off-chain simulation of what the account's state *would be* if a settlement were to happen at the current block, without actually making any state changes.
 
-- **Parameters**:
-  - `token`: ERC20 token contract address
-  - `owner`: Account address being queried
```
rvagg, line 241: I'd like these back, but you could put them down with the return values and just remove the "Understanding the" there, they're basically this original list of return values but explained better

When I look at API docs, the thing I want front and center are arguments and return values, secondarily any error conditions I should be aware of; prose is fine as an introduction but it doesn't help me implement and "look at the source bro" isn't a great answer.

rvagg: which does argue against using elipses to truncate long argument strings, the closer you are to presenting an ABI, the easier it is for tooling to just use it directly - imagine pointing Claude at this README and telling it to go inspect contract state with `cast`

README.md
```diff
@@ -48,298 +48,400 @@ The FWS Payments contract enables ERC20 token payment flows through "rails" - au
...
 #### `getRail(uint256 railId)`
 
 Retrieves the current state of a payment rail.
+
+- **When to use:** To inspect the parameters of an existing rail.
 - **Parameters**:
-  - `railId`: Rail identifier
-- **Returns**: RailView struct with rail details
+  - `railId`: The rail's unique identifier.
+- **Returns**: A `RailView` struct containing the rail's public data.
```
rvagg, line 328: I'd love to see a `RailView` in here and have it explained

README.md
```diff
@@ -48,298 +48,400 @@ The FWS Payments contract enables ERC20 token payment flows through "rails" - au
...
+-   **When to use:** An operator calls this to adjust the client's funding guarantee. This is used to set an initial `lockupFixed` for an onboarding fee, increase the `lockupPeriod` for a longer-term commitment, or decrease lockups when a deal's terms change.
+
+**Lockup Calculation and State Changes**
+
+This function recalculates the rail's total lockup requirement based on the new `period` and `lockupFixed` values. The change in the rail's individual lockup is then applied to the client's total account lockup (`Account.lockupCurrent`).
+
+-   **State Impact:** It modifies both the `Rail` struct (updating `lockupPeriod` and `lockupFixed`) and the client's `Account` struct (updating `lockupCurrent`).
+-   **Safety Checks:** The entire operation is wrapped in a modifier that settles the client's account lockup before and after the change. This ensures that any increase in lockup is only permitted if the client has sufficient available funds to cover it, maintaining the contract's solvency invariants.
+
+-   **Parameters**:
```
rvagg, line 378: not sure these should be `-` dotpoints; maybe just lift them up as top-level things, otherwise they look like they are under "Lockup Caluclation and State Changes", same with `modifyRailPayment` below and `terminateRail` above

README.md
```diff
@@ -650,34 +757,48 @@ Payments(paymentsContractAddress).modifyRailLockup(
...
 
-When a user cancels a deal, the service contract will modify the rail's payment to take that into account. In this case, the service contract sets the rail's payment rate to zero and pays a fixed termination fee out of the rail's "fixed" lockup.
+There are two primary ways to end a deal:
+
+**Method 1: Soft End (Rate to Zero)**
+The operator can set the payment rate to zero and optionally charge a final termination fee. This keeps the rail active but stops recurring payments.
 
 ```solidity
-// Service contract reduces payment rate and possibly issues a termination payment
+// Service contract reduces payment rate and issues a termination payment
```
rvagg, line 768: above you said "optionally charge a final termination fee" - is that the same thing as "termination payment" and therefore should that optionality be reflected in the language here?

README.md
aarshkshah1992: @ZenGround0 Not sure what you mean, The new bullet list has everything the old list has but in more detail

README.md
```diff
@@ -48,298 +48,400 @@ The FWS Payments contract enables ERC20 token payment flows through "rails" - au
...
 
 ### Operator
 
-An operator is a smart contract (typically the "service contract" for a given service) that manages rails on behalf of clients & service providers, with approval from the client (the client approves the operator to spend its funds at a specific rate). The operator smart contract must be trusted by both the client and the service provider as it can arbitrarily alter payments (within the allowance specified by the client).
+An operator is a smart contract—typically the main contract for a given service—that manages payment rails on behalf of clients. A client must explicitly approve an operator and grant it specific allowances, which act as a budget for how much the operator can spend or lock up on the client's behalf.
 
-The operator:
+The operator role is powerful, so the operator contract must be trusted by both the client and the service provider. The client trusts it not to abuse its spending allowances, and the service provider trusts it to correctly configure and manage the payment rail.
```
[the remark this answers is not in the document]
aarshkshah1992: Done

README.md
```diff
@@ -1,14 +1,14 @@
...
+# FWS Payments Contract: A Builder's Manual
 
-The FWS Payments contract enables ERC20 token payment flows through "rails" - automated payment channels between clients and recipients. The contract supports continuous payments, one-time transfers, and payment validation / arbitration.
+The FWS Payments contract enables ERC20 token payment flows through "rails" - automated payment channels between clients and recipients. The contract supports continuous rate based payments, one-time transfers, and payment validation during settlement.
 
 - [Deployment Info](#deployment-info)
 - [Key Concepts](#key-concepts)
   - [Account](#account)
   - [Rail](#rail)
   - [Validator](#validator)
```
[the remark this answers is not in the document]
aarshkshah1992: Yes 

- `Validator` and `Verifier` are the same things. But let's call it the `Validator`
- Service contract can be the `Validator` as well but does not have to be
- Service contract can configure whatever `Validator` it prefers.

README.md
```diff
@@ -48,298 +48,400 @@ The FWS Payments contract enables ERC20 token payment flows through "rails" - au
...
-  - The sum of the lockup across all rails operated by this contract for a specific client must be at most the maximum per-operator lockup specified by the client.
-- Specifies / changes the payment rail validator of payment rails created by this operator.
-- Terminates payment rails.
+An approved operator can perform the following actions:
 
-### Per-Rail Lockup: Streaming and Fixed Buckets
+-   **Create Rails (`createRail`):** Establish a new payment rail from a client to a service provider, specifying the token, payee, and an optional validator.
+-   **Modify Rail Terms (`modifyRailLockup`, `modifyRailPayment`):** Adjust the payment rate, lockup period, and fixed lockup amount for any rail it manages. Any increase in the client's financial commitment is checked against the operator's allowances.
+-   **Execute One-Time Payments (`modifyRailPayment`):** Execute one-time payments from the rail's fixed lockup.
+-   **Settle Rails (`settleRail`):** Trigger payment settlement for a rail to process due payments. As a rail participant, the operator can initiate settlement at any time.
```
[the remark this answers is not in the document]
aarshkshah1992: Done.

README.md
aarshkshah1992: @rvagg No strong opinions though I like "Streaming Lockup". But we can keep this open till we have more feedback from the outside world.

README.md
aarshkshah1992: What I want to say is that you can not draw from this "streaming lockup" in the normal course of action for providing your service. You can ONLY use it after you terminate a rail and so it is a guarantee of solvency for a pre-agreed number of epochs. Let me add that to the framing.

We've already explained all over the doc that a client/payer can NEVER withdraw locked funds.

README.md
aarshkshah1992: Done.

README.md
```diff
@@ -48,298 +48,400 @@ The FWS Payments contract enables ERC20 token payment flows through "rails" - au
...
-- **Use cases:** Use streaming lockup for predictable, recurring payments. Use fixed lockup for immediate payments.
-
-This model gives flexibility: this can guarantee ongoing service payments while also handling special, immediate payments as needed—all with clear, enforced separation and safety for both parties.
+-   **Settle Regularly:** You must settle rails frequently. A rail's `lockupPeriod` is a measure of the risk you are willing to take. If you wait longer than the `lockupPeriod` to settle, you allow a client to build up a payment obligation that may not be fully covered by the lockup guarantee if they become insolvent.
+-   **Monitor Client Solvency:** Use the `getAccountInfoIfSettled` function to check if a client is funded. If their `fundedUntilEpoch` is approaching the current epoch, they are at risk.
+-   **Terminate Proactively:** If a client becomes insolvent or unresponsive, terminate the rail immediately. This is the **only way** to activate the safety hatch and ensure you can claim payment from the funds guaranteed by the streaming lockup.
```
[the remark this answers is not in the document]
aarshkshah1992: nice catch. Done.

README.md
aarshkshah1992: Agreed, adding these back for all.

README.md
```diff
@@ -48,298 +48,400 @@ The FWS Payments contract enables ERC20 token payment flows through "rails" - au
...
 
-- **Parameters**:
-  - `token`: ERC20 token contract address
-  - `owner`: Account address being queried
-- **Returns**:
-  - `fundedUntilEpoch`: epoch until which account is fully funded
-  - `currentFunds`: currently available funds before settling
-  - `availableFunds`: funds available if settlement were to happen now, clamped at 0
-  - `currentLockupRate`: the current lockup rate per epoch
+**Why and How to Use This Function**
```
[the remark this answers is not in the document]
aarshkshah1992: Done

README.md
aarshkshah1992: Good point. Done.

README.md
```diff
@@ -437,36 +539,42 @@ A client can reduce the operator's `maxLockupPeriod` or `lockupAllowance` after
 
 ### Settlement
 
+Functions for processing payments by moving funds from the client to the payee based on the rail's terms.
+
 #### `settleRail(uint256 railId, uint256 untilEpoch)`
```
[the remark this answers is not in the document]
aarshkshah1992: Fixed.

README.md
```diff
@@ -437,36 +539,42 @@ A client can reduce the operator's `maxLockupPeriod` or `lockupAllowance` after
...
+
+**Note**: While the validator has significant control, the final settlement outcome is also dependent on the client having sufficient funds for the amount being settled.
 
 #### `settleTerminatedRailWithoutValidation(uint256 railId)`
 
-Emergency settlement method for terminated rails with stuck validation.
+This is a crucial escape-hatch function that allows the **client** to finalize a terminated rail that is otherwise stuck, for example, due to a malfunctioning validator.
 
-- **Parameters**:
-  - `railId`: Rail identifier
```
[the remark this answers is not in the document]
aarshkshah1992: Fixed.

README.md
aarshkshah1992: Good catch. Done.

README.md
aarshkshah1992: Done.

README.md
aarshkshah1992: I have linked the "ONE TIME Payments" section here.

aarshkshah1992:
@rvagg @ZenGround0 I have addressed the comments. Please can you give it one more look ?

.github/scripts/check-contract-size.sh
```diff
@@ -36,6 +36,13 @@ if ! jq empty contract_sizes.json 2>/dev/null; then
     exit 1
 fi
 
+if jq -e '. == {}' contract_sizes.json >/dev/null; then
+    echo "forge did not find any contracts. forge build:"
+    # This usually means build failure
+    forge build
+    exit 1
+fi
+
```
aarshkshah1992, line 45: why are these changes flowing in ? Let me check.

README.md
```diff
@@ -47,299 +47,477 @@ The FWS Payments contract enables ERC20 token payment flows through "rails" - au
...
+**1. During Normal Operation (Before Termination)**
 
-- You must ensure both components are sufficiently funded.
-- One-time payments are deducted from `lockupFixed`.
-- Streaming payments are covered by the streaming lockup.
+While a rail is active, the streaming lockup acts as a **guarantee of solvency for a pre-agreed number of epochs**, not as a direct source of payment.
 
-#### How Each Lockup Bucket Works
+-   **Payments from General Funds:** When `settleRail` is called on an active rail, payments are drawn from the payer's general `funds`.
+-   **Lockup as a Floor:** The lockup simply acts as a minimum balance. The contract prevents the payer from withdrawing funds below this floor.
```
pyropy, line 114: In case we want to have a pre-payment type of service then we'd need to:

- Set lockup fixed to some N amount of tokens
- Have user deposit N amount of tokens to their account, updating the account `funds` to equal the deposited amount 

Is this correct?

pyropy (commented):
Great job @aarshkshah1992! I found the guide really helpful. 

I think it would be also a good idea to dedicate a section that outlines how to create different payment models (i.e. resource-usage payment model). That section would probably also expand a bit on lockups as in that payment model you can really rely on the streaming lockup (due the rate being unpredictable).

rvagg (approved):
I haven't done a second pass on this, but my first pass was generally 👍 with some minor nits. Overall I think this is a great improvement and we're better off having this merged and fixing up any problems with follow-ups than letting it sit in a branch.

README.md
aarshkshah1992: @pyropy The call to set the fixed lockup i.e. `modifyTerminatedRailLockup` will fail if the user does not already enough funds to be locked.

aarshkshah1992:
Merging this as all comments have been addressed.

PR #204, opened 2025-08-07
Block Timestamp

wjmelements:
Reviewer @ZenGround0 @rvagg
The time between blocks might change in the future.
This should not impact the duration of agreements or the payment rate over time.
#### Changes
* replace block.number with block.timestamp
* TODO rename epoch

### DIFF
--- a/README.md
+++ b/README.md
@@ -617,7 +617,7 @@ Payment settlement can be triggered by any rail participant:
 // Settlement call - can be made by client, service provider, or operator
 (uint256 amount, uint256 settledEpoch, string memory note) = Payments(paymentsContractAddress).settleRail(
     railId,        // Rail ID
-    block.number   // Settle up to current epoch
+    block.timestamp   // Settle up to now
 );
 ```
 
--- a/src/Payments.sol
+++ b/src/Payments.sol
@@ -693,7 +693,7 @@ contract Payments is ReentrancyGuard {
         rail.to = to;
         rail.operator = operator;
         rail.validator = validator;
-        rail.settledUpTo = block.number;
+        rail.settledUpTo = block.timestamp;
         rail.endEpoch = 0;
         rail.commissionRateBps = commissionRateBps;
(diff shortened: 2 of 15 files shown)


README.md
```diff
@@ -617,7 +617,7 @@ Payment settlement can be triggered by any rail participant:
 // Settlement call - can be made by client, service provider, or operator
 (uint256 amount, uint256 settledEpoch, string memory note) = Payments(paymentsContractAddress).settleRail(
     railId,        // Rail ID
-    block.number   // Settle up to current epoch
+    block.timestamp   // Settle up to now
```
wjmelements, line 620: Specifying a current timestamp parameter might be harder than specifying a block number. Two solutions to this:
1. `min(block.timestamp, untilTime)`
2. add helper to settle up to now.

src/Payments.sol
```diff
@@ -846,8 +846,8 @@ contract Payments is ReentrancyGuard {
         if (isTerminated) {
             uint256 maxSettlementEpoch = maxSettlementEpochForTerminatedRail(rail, railId);
             require(
-                block.number < maxSettlementEpoch,
-                Errors.CannotModifyTerminatedRailBeyondEndEpoch(railId, maxSettlementEpoch, block.number)
+                block.timestamp < maxSettlementEpoch,
```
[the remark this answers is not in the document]
wjmelements: Both are timestamps in this situation. I haven't yet renamed epoch, per the PR description.

README.md
wjmelements: > 1. `min(block.timestamp, untilTime)`
> 2. add helper to settle up to now.


For UX, I think I want to do one of these whether or not we make this change.

aarshkshah1992:
See discussion at https://filecoinproject.slack.com/archives/C07CGTXHHT4/p1754585863232389.

wjmelements:
Closed because we don't expect the block time to change anytime soon and because the change would be difficult.

PR #207, opened 2025-08-08
Profile FVM Gas With Local Lotus

wjmelements:
The current approach is to use `forge script --broadcast`.
[Rationale](https://github.com/FilOzone/forge-script-gas-report?tab=readme-ov-file#why-not-use-forge-snapshot)
The profile script covers some of the basic Payment operations.
The scripts are run sequentially to ensure that time has passed.
#### Changes
* add Profile.s.sol and profile.sh
* make `.gas-report` with https://github.com/FilOzone/forge-script-gas-report
* eliminate gas variance
* TODO compare [localnet](https://github.com/Filozone/localnet) to itest/ensemble approach
* TODO gas-profile GH Action

Issue #205, opened 2025-08-08
Add ERC-3009 Support


rjan90:
## Background
USDFC gained [ERC-3009](https://eips.ethereum.org/EIPS/eip-3009) superpowers on Calibnet

## Work
- Add ERC-3009 transferWithAuthorization/receiveWithAuthorization support
- Create deposit methods accepting ERC-3009 authorization parameters
- Create an "uber" method similar to depositWithPermitAndApproveOperator that combines ERC-3009 authorized transfer with operator approval in a single transaction
- Add test coverage

## Context:
This builds upon our existing EIP-2612 implementation and the [depositWithPermitAndApproveOperator](https://github.com/FilOzone/filecoin-pay/blob/18b220a65af5fc2d060ab4793201687cc7689191/src/Payments.sol#L558-L559) pattern to provide users with ERC-3009's benefits (random nonces, better UX, reduced front-running risk).

pali101:
This looks interesting. Can I pick it up @rjan90?

rjan90:
@pali101 ofcourse! Will assign it to you

Chaitu-Tatipamula:
Hey @pali101 are you working on this? if not i would love to pick this up!

pali101:
> Hey [@pali101](https://github.com/pali101) are you working on this? if not i would love to pick this up!

Hi @Chaitu-Tatipamula, I've already started working on it.

wjmelements:
> I think the 3009 nonce should be a hash of an EIP-712 structure

One danger to this is that the client isn't necessarily aware that our ERC3009 `nonce` encodes this information. Their signature prompt is likely to hide such allowance increases. In Permit2 this is more explicit (`witness`). I currently believe this is fine. We can make sure that official client software surfaces this information to limit attack vectors.

wjmelements:
> > I think the 3009 nonce should be a hash of an EIP-712 structure
> 
> One danger to this is that the client isn't necessarily aware that our ERC3009 `nonce` encodes this information. Their signature prompt is likely to hide such allowance increases. In Permit2 this is more explicit (`witness`). I currently believe this is fine. We can make sure that official client software surfaces this information to limit attack vectors.

I went to the discussions-to for erc3009 and found the thread was locked. I have contacted pandapip1 to possibly unlock the thread.

PR #214, opened 2025-08-20
feat: add support for ERC-3009

pali101:
## Summary

Adds `ERC-3009` deposit flows, event extensions, and tests. 

## Changes

- Introduce `IERC3009`.
- Payments
    - Add `AuthType` enum; extend `DepositRecorded(token, from, to, amount, authType, nonce)`.
    - Generalize self-recipient check: `PermitRecipientMustBeMsgSender` → `SignerMustBeMsgSender`; `validatePermitRecipient` → `validateSignerIsRecipient`.
    - New methods: `depositWithAuthorization(...)`, `depositWithAuthorizationAndApproveOperator(...)`, `depositWithAuthorizationAndIncreaseOperatorApproval(...)`
- Tests
     - Comprehensive coverage for happy path, replay, signature errors, time windows, sender mismatch, insufficient balance, domain mismatch.
- Mock
     - `MockERC20`: implements ERC-3009. 

## Notes

- `authorizationState` used for replay protection
- `domain separation` enforced by signatures.

### DIFF
--- a/src/Payments.sol
+++ b/src/Payments.sol
@@ -33,6 +33,32 @@ interface IValidator {
     function railTerminated(uint256 railId, address terminator, uint256 endEpoch) external;
 }
 
+interface IERC3009 {
+    /**
+     * @notice Execute a transfer with a signed authorization
+     * @param from          Payer's address (Authorizer)
+     * @param to            Payee's address
+     * @param value         Amount to be transferred
+     * @param validAfter    The time after which this is valid (unix time)
+     * @param validBefore   The time before which this is valid (unix time)
+     * @param nonce         Unique nonce
+     * @param v             v of the signature
+     * @param r             r of the signature
+     * @param s             s of the signature
+     */
+    function transferWithAuthorization(
+        address from,
+        address to,
+        uint256 value,
+        uint256 validAfter,
+        uint256 validBefore,
+        bytes32 nonce,
+        uint8 v,
+        bytes32 r,
+        bytes32 s
+    ) external;
+}
+
 // @title Payments contract.
 contract Payments is ReentrancyGuard {
     using SafeERC20 for IERC20;
@@ -43,6 +69,12 @@ contract Payments is ReentrancyGuard {
 
     uint256 public constant NETWORK_FEE = 1300000 gwei; // equivalent to 130000 nFIL
     address payable private constant BURN_ADDRESS = payable(0xff00000000000000000000000000000000000063);
+
+    enum AuthType {
+        None,
+        Permit,
+        Authorization
+    }
 
     // Events
     event AccountLockupSettled(
@@ -92,7 +124,12 @@ contract Payments is ReentrancyGuard {
     event RailFinalized(uint256 indexed railId);
 
     event DepositRecorded(
-        address indexed token, address indexed from, address indexed to, uint256 amount, bool usedPermit
+        address indexed token,
+        address indexed from,
+        address indexed to,
+        uint256 amount,
+        AuthType authType,
+        bytes32 nonce
     );
     event WithdrawRecorded(address indexed token, address indexed from, address indexed to, uint256 amount);
 
@@ -475,7 +512,7 @@ contract Payments is ReentrancyGuard {
 
         account.funds += actualAmount;
 
-        emit DepositRecorded(token, msg.sender, to, actualAmount, false);
+        emit DepositRecorded(token, msg.sender, to, actualAmount, AuthType.None, bytes32(0));
     }
 
     /**
@@ -530,7 +567,7 @@ contract Payments is ReentrancyGuard {
 
         account.funds += actualAmount;
 
-        emit DepositRecorded(token, to, to, actualAmount, true);
+        emit DepositRecorded(token, to, to, actualAmount, AuthType.Permit, bytes32(0));
     }
 
     /**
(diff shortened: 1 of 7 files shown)


src/Payments.sol
```diff
@@ -606,13 +647,162 @@ contract Payments is ReentrancyGuard {
...
+        uint256 lockupAllowanceIncrease
+    )
+        external
+        nonReentrant
+        validateNonZeroAddress(operator, "operator")
+        validateNonZeroAddress(to, "to")
+        validateSignerIsRecipient(to)
+        settleAccountLockupBeforeAndAfter(token, to, false)
+    {
+        _increaseOperatorApproval(token, operator, rateAllowanceIncrease, lockupAllowanceIncrease);
```
wjmelements, line 769: [Previous comment](https://github.com/FilOzone/filecoin-pay/issues/205#issuecomment-3184661905)

Our goal with supporting ERC 3009 is to allow the client to deposit and increase approval with one signature. This is two:
* ERC3009 permit
* Filecoin transaction

Currently `_increaseOperatorApproval` and `_setOperatorApproval` use `msg.sender`, so the sender must be the permitter and this means the client has to sign twice. This is necessary because the ERC3009 permit signs for the deposit and the Filecoin transaction signs for the `operator`, `rateAllowanceIncrease`, and `lockupAllowanceIncrease`.

The way we can combine these into one signature is to use the ERC3009 nonce:
```
bytes32 constant INCREASE_ALLOWANCE_TYPEHASH = keccak256("IncreaseAllowance(address operator,uint256 rateAllowanceIncrease,uint256 lockupAllowanceIncrease)");
bytes32 erc3009Nonce = keccak256(abi.encode(INCREASE_ALLOWANCE_TYPEHASH, operator, rateAllowanceIncrease, lockupAllowanceIncrease, nonce));
```
The core idea is to encapsulate the full intent into one signature containing both the permit and the allowance increase. Thereby we should be able to remove the `msg.sender` restrictions for the erc3009 deposit methods. For `depositWithAuthorization`, `rateAllowanceIncrease` and `lockupAllowanceIncrease` should be zero.

test/mocks/MockERC20.sol
```diff
@@ -2,12 +2,170 @@
...
+
+    bytes32 private _HASHED_NAME = keccak256("USD for Filecoin");
+    bytes32 private _HASHED_VERSION = keccak256("1");
+
+    // keccak256("Permit(address owner,address spender,uint256 value,uint256 nonce,uint256 deadline)");
+    bytes32 private constant _PERMIT_TYPEHASH = 0x6e71edae12b1b97f4d1f60370fef10105fa2faae0126114a169c64845d6126c9;
+    // keccak256("EIP712Domain(string name,string version,uint256 chainId,address verifyingContract)");
+    bytes32 private constant _TYPE_HASH = 0x8b73c3c69bb8fe3d512ecc4cf759cc79239f7b179b0ffacaa9a75d522b39400f;
+
+    uint256 private _CACHED_CHAIN_ID = _chainID();
```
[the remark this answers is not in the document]
wjmelements: The consideration for storing `_chainID` vs `block.chainId` is that hard forks can change `block.chainid`. This is a mock test token so it doesn't especially matter.

src/Payments.sol
```diff
@@ -498,7 +539,7 @@ contract Payments is ReentrancyGuard {
         external
         nonReentrant
         validateNonZeroAddress(to, "to")
-        validatePermitRecipient(to)
+        validateSignerIsRecipient(to)
```
wjmelements, line 542: I don't think we need this `msg.sender` check for a permit deposit. Removing this would cut deposit signatures from 2 to 1.

wjmelements: Yes but the tx signature doesn't have to be the same user and so such txs can batch these deposits with other deposits or even other operations. In particular if we do [this](https://github.com/FilOzone/filecoin-pay/pull/214#discussion_r2289273100) the operator is incentivized to complete your deposit.

pali101: That makes a lot of sense. It would definately add a lot of value in long term. 

I recall that @rvagg ran into an issue when trying to make our implementation work with deposits from a different address (see #142). If we plan to support submissions from external addresses, I can look into the root cause of that problem.

src/Payments.sol
```diff
@@ -498,7 +526,7 @@ contract Payments is ReentrancyGuard {
...
```
wjmelements, line 529: We also want to remove this for 2612.

src/Payments.sol
```diff
@@ -606,13 +634,156 @@ contract Payments is ReentrancyGuard {
...
+     * @param validBefore The timestamp before which the authorization is valid.
+     * @param nonce A unique nonce for the authorization, used to prevent replay attacks.
+     * @param v,r,s The signature of the authorization.
+     * @param operator The address of the operator whose allowances are being increased.
+     * @param rateAllowanceIncrease The amount to increase the rate allowance by.
+     * @param lockupAllowanceIncrease The amount to increase the lockup allowance by.
+     * @custom:constraint Operator must already be approved.
+     */
+    function depositWithAuthorizationAndIncreaseOperatorApproval(
+        address token,
```
wjmelements, line 730: IERC3009 token

src/Payments.sol
```diff
@@ -33,6 +33,36 @@ interface IValidator {
     function railTerminated(uint256 railId, address terminator, uint256 endEpoch) external;
 }
 
+interface IERC3009 {
```
wjmelements, line 36: IERC3009 is ERC20

test/mocks/MockERC20.sol
```diff
@@ -2,12 +2,168 @@
 pragma solidity ^0.8.27;
 
 import "@openzeppelin/contracts/token/ERC20/extensions/ERC20Permit.sol";
+import "@openzeppelin/contracts/utils/cryptography/ECDSA.sol";
 
+/**
+ * @title MockERC20
+ * @dev A mock ERC20 token with permit (ERC-2612) and transferWithAuthorization (ERC-3009) functionality for testing purposes.
+ */
 contract MockERC20 is ERC20Permit {
```
wjmelements, line 11: also is erc3009, erc2612

wjmelements: ok I see oz calls erc2612 erc20permit

src/Payments.sol
```diff
@@ -606,13 +634,156 @@ contract Payments is ReentrancyGuard {
...
+        // Revert if token is address(0) as authorization is not supported for native tokens
+        require(token != address(0), Errors.NativeTokenNotSupported());
+
+        // Use balance-before/balance-after accounting to correctly handle fee-on-transfer tokens
+        uint256 balanceBefore = IERC20(token).balanceOf(address(this));
+
+        // Call ERC-3009 transferWithAuthorization.
+        // This will transfer 'amount' from 'to' to this contract.
+        // The token contract itself verifies the signature.
+        IERC3009(token).receiveWithAuthorization(to, address(this), amount, validAfter, validBefore, nonce, v, r, s);
```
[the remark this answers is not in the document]
pali101: ERC3009 `receiveWithAuthorization` function does not return a value. It either completes successfully or reverts on failure so requiring a return is not applicable.

Reference: https://github.com/Secured-Finance/stablecoin-contracts/blob/develop/contracts/DebtToken.sol#L288-L325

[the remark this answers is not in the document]
pali101: I reviewed the official ERC-3009 specification, which defines the function as follows:

```solidity
function receiveWithAuthorization(
    address from,
    address to,
    uint256 value,
    uint256 validAfter,
    uint256 validBefore,
    bytes32 nonce,
    uint8 v,
    bytes32 r,
    bytes32 s
) external;
```

As you can see, the function signature does not specify any return value. If USDFC's implementation were to return a value, this would alter the function signature and thus break compliance with the ERC-3009 standard.

Reference: https://eips.ethereum.org/EIPS/eip-3009#:~:text=/**%0A%20*%20%40notice%20Receive%20a,)%20external%3B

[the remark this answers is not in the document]
wjmelements: You are right. I must have confused this somehow.

src/Payments.sol
pali101: Made that change in separate PR - #217

test/mocks/MockERC20.sol
pali101: Done, it now inherits ERC20, ERC20Permit (i.e., ERC-2612) and implements IERC3009.

wjmelements:
I wonder why your merge looks like this

Screenshot 2025-09-05 at 2 58 56 PM

Normally Github only shows the merge commit, not all of the commits in the other branch. [Example](https://github.com/FilOzone/filecoin-pay/pull/207) How did you merge main in git?

Screenshot 2025-09-05 at 3 00 13 PM

src/Payments.sol
```diff
@@ -600,13 +598,156 @@ contract Payments is ReentrancyGuard {
...
+     *             on behalf of the message sender. If this is less than the current payment rate, the operator will
+     *             only be able to reduce rates until they fall below the target.
+     * @param lockupAllowance The maximum amount of funds the operator can lock up on behalf of the message sender
+     *             towards future payments. If this exceeds the current total amount of funds locked towards future payments,
+     *             the operator will only be able to reduce future lockup.
+     * @param maxLockupPeriod The maximum number of epochs (blocks) the operator can lock funds for. If this is less than
+     *             the current lockup period for a rail, the operator will only be able to reduce the lockup period.
+     */
+    function depositWithAuthorizationAndApproveOperator(
+        address token,
```
wjmelements, line 653: IERC3009

rjan90:
@pali101 It seems like there is only three small nits that you need to address here, before we can merge this PR 🙏

PR #219, opened 2025-09-08
Anyone Can settleRail

wjmelements:
Reviewer @aarshkshah1992
By allowing anyone to settleRail, realizable fee growth will be smoother, reducing MEV.
#### Changes
* remove onlyRailParticipant check from settelRail

### DIFF
--- a/src/Payments.sol
+++ b/src/Payments.sol
@@ -202,14 +202,6 @@ contract Payments is ReentrancyGuard {
         _;
     }
 
-    modifier onlyRailParticipant(uint256 railId) {
-        require(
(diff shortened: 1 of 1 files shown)


rvagg:
Is there any potential tax reason, in some jurisdiction, that someone may not want to have a rail settled until a particular time?
This change seems entirely reasonable to me, just trying to think through any negative consequences for participants.

wjmelements:
> in some jurisdiction

Probably somewhere in the world, but checks are taxable income when received, not when cashed or deposited. Most jurisdictions don't let you defer income except in specific circumstances.

PR #225, opened 2025-09-16
IERC20 type, and label mapping keys

wjmelements:
Reviewer @rvagg
Labeling the mapping keys improves the ABI by naming the parameters and helps document the code.
Correct types reduce casting, catch a swapped-param bug in our events, and don't change the ABI.
`interface` types are `address` in the function signature.

After this merges I'll put up a similar changeset in filecoin-services.

#### Changes
* token is IERC20, not address
* label mapping keys
* fix event where args are flipped

### DIFF
--- a/src/Payments.sol
+++ b/src/Payments.sol
@@ -365,7 +365,7 @@ contract Payments is ReentrancyGuard {
         uint256 rateAllowanceIncrease,
         uint256 lockupAllowanceIncrease
     ) external nonReentrant validateNonZeroAddress(operator, "operator") {
-        _increaseOperatorApproval(IERC20(token), operator, rateAllowanceIncrease, lockupAllowanceIncrease);
+        _increaseOperatorApproval(token, operator, rateAllowanceIncrease, lockupAllowanceIncrease);
     }
 
     function _increaseOperatorApproval(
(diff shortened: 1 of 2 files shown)


wjmelements:
I still need to fix tests but my computer is being slow
> [⠆] Solc 0.8.30 finished in 498.23s

src/Errors.sol
```diff
@@ -109,15 +111,15 @@ library Errors {
     /// @param token The token involved in the lockup
     /// @param currentLockup The payer's current lockup amount
     /// @param lockupReduction The reduction attempted to be made
-    error InsufficientCurrentLockup(address from, address token, uint256 currentLockup, uint256 lockupReduction);
+    error InsufficientCurrentLockup(IERC20 token, address from, uint256 currentLockup, uint256 lockupReduction);
```
wjmelements, line 114: These parameters were switched and it was a bug. Auditors missed it.

wjmelements: > These parameters were switched and it was a bug. Auditors missed it.

found when changing token to IERC20:
```
Error (9553): Invalid type for argument in function call. Invalid implicit conversion from contract IERC20 to address requested.
   --> src/Payments.sol:892:46:
    |
892 |             Errors.InsufficientCurrentLockup(rail.token, rail.from, payer.lockupCurrent, lockupReduction)
    |                                              ^^^^^^^^^^

Error (9553): Invalid type for argument in function call. Invalid implicit conversion from address to contract IERC20 requested.
   --> src/Payments.sol:892:58:
    |
892 |             Errors.InsufficientCurrentLockup(rail.token, rail.from, payer.lockupCurrent, lockupReduction)
    |                                                          ^^^^^^^^^
```

rvagg (approved):
from what I can tell this doesn't break the ABI, just that event and solidity-level API, please correct me if I'm wrong

Issue #218, opened 2025-09-03
Payments can implement ERC6909


wjmelements:
ERC6909 is a multitoken standard appropriate for internal balances. If we implement this standard, deposits could be more visible to the user, if a wallet or explorer supports ERC6909.

wjmelements:
This change would also help with the datacap allocator.

pali101:
@rjan90 Please assign it to me.

PR #196, opened 2025-07-31
rm Proxy and Ownable

wjmelements:
Closes https://github.com/FilOzone/filecoin-services-payments/issues/195
Reviewer @ZenGround0

I also noticed that the only code using Ownable was the proxy so this also removes Ownable.

#### Changes
* rm proxy
* change libraries to non proxy versions
* move initialize to constructor
* remove Ownable
* update tools/ and Makefile

PR #229, opened 2025-09-16
Fee Auction

wjmelements:
Reviewers @rvagg @aarshkshah1992
#### Changes
* remove fixed fee settlement mechanism
* collect fees into payments account
* auction fees
* TODO handle fee-on-transfer tokens
* TODO adaptive auction price
* TODO auction tests

### DIFF
--- a/src/Payments.sol
+++ b/src/Payments.sol
@@ -42,8 +42,8 @@ contract Payments is ReentrancyGuard {
     uint256 public constant COMMISSION_MAX_BPS = 10000;
     uint256 private constant AUCTION_START_PRICE = .083 ether; // 0.083 FIL
 
-    uint256 private constant NETWORK_FEE_NUMERATOR = 5; // 0.5%
-    uint256 private constant NETWORK_FEE_DENOMINATOR = 1000;
+    uint256 public constant NETWORK_FEE_NUMERATOR = 5; // 0.5%
+    uint256 public constant NETWORK_FEE_DENOMINATOR = 1000;
 
     address payable private constant BURN_ADDRESS = payable(0xff00000000000000000000000000000000000063);
     IERC20 private constant NATIVE_TOKEN = IERC20(address(0));
@@ -83,13 +83,14 @@ contract Payments is ReentrancyGuard {
         uint256 oldLockupFixed,
         uint256 newLockupFixed
     );
-    event RailOneTimePaymentProcessed(uint256 indexed railId, uint256 netPayeeAmount, uint256 operatorCommission);
+    event RailOneTimePaymentProcessed(uint256 indexed railId, uint256 netPayeeAmount, uint256 operatorCommission, uint256 networkFee);
     event RailRateModified(uint256 indexed railId, uint256 oldRate, uint256 newRate);
     event RailSettled(
         uint256 indexed railId,
         uint256 totalSettledAmount,
         uint256 totalNetPayeeAmount,
         uint256 operatorCommission,
+        uint256 networkFee,
         uint256 settledUpTo
     );
     event RailTerminated(uint256 indexed railId, address indexed by, uint256 endEpoch);
@@ -1120,7 +1121,7 @@ contract Payments is ReentrancyGuard {
             // Credit payee (net amount after fees)
             payee.funds += netPayeeAmount;
 
-            emit RailOneTimePaymentProcessed(railId, netPayeeAmount, operatorCommission);
+            emit RailOneTimePaymentProcessed(railId, netPayeeAmount, operatorCommission, networkFee);
         }
     }
 
@@ -1240,7 +1241,7 @@ contract Payments is ReentrancyGuard {
         finalSettledEpoch = rail.settledUpTo;
         note = checkAndFinalizeTerminatedRail(railId, rail, payer, note);
 
-        emit RailSettled(railId, totalSettledAmount, totalNetPayeeAmount, totalOperatorCommission, finalSettledEpoch);
+        emit RailSettled(railId, totalSettledAmount, totalNetPayeeAmount, totalOperatorCommission, totalNetworkFee, finalSettledEpoch);
 
         return (totalSettledAmount, totalNetPayeeAmount, totalOperatorCommission, totalNetworkFee, finalSettledEpoch, note);
     }
--- a/src/Payments.sol
+++ b/src/Payments.sol
@@ -1073,7 +1073,12 @@ contract Payments is ReentrancyGuard {
         returns (uint256 netPayeeAmount, uint256 operatorCommission, uint256 fee)
     {
         fee = (amount * NETWORK_FEE_NUMERATOR + (NETWORK_FEE_DENOMINATOR - 1)) / NETWORK_FEE_DENOMINATOR;
-        accounts[token][address(this)].funds += fee;
+        if (token == NATIVE_TOKEN) {
+            (bool success,) = BURN_ADDRESS.call{value: fee}("");
+            require(success, Errors.NativeTransferFailed(BURN_ADDRESS, msg.value));
+        } else {
+            accounts[token][address(this)].funds += fee;
+        }
         amount -= fee;
 
         // Calculate operator commission (if any) based on remaining amount
--- a/src/Payments.sol
+++ b/src/Payments.sol
@@ -41,7 +41,10 @@ contract Payments is ReentrancyGuard {
     // Maximum commission rate in basis points (100% = 10000 BPS)
     uint256 public constant COMMISSION_MAX_BPS = 10000;
     uint256 private constant AUCTION_START_PRICE = .083 ether; // 0.083 FIL
-    uint256 public constant NETWORK_FEE = 1300000 gwei; // equivalent to 1300000 nanoFIL / 0.0013 FIL
+
+    uint256 private constant NETWORK_FEE_NUMERATOR = 5; // 0.5%
+    uint256 private constant NETWORK_FEE_DENOMINATOR = 1000;
+
     address payable private constant BURN_ADDRESS = payable(0xff00000000000000000000000000000000000063);
     IERC20 private constant NATIVE_TOKEN = IERC20(address(0));
 
@@ -177,6 +180,7 @@ contract Payments is ReentrancyGuard {
         uint256 totalSettledAmount;
         uint256 totalNetPayeeAmount;
         uint256 totalOperatorCommission;
+        uint256 totalNetworkFee;
         uint256 processedEpoch;
         string note;
     }
@@ -1066,8 +1070,12 @@ contract Payments is ReentrancyGuard {
 
     function calculateAndPayFees(uint256 amount, IERC20 token, address serviceFeeRecipient, uint256 commissionRateBps)
         internal
-        returns (uint256 netPayeeAmount, uint256 operatorCommission)
+        returns (uint256 netPayeeAmount, uint256 operatorCommission, uint256 fee)
     {
+        fee = (amount * NETWORK_FEE_NUMERATOR + (NETWORK_FEE_DENOMINATOR - 1)) / NETWORK_FEE_DENOMINATOR;
+        accounts[token][address(this)].funds += fee;
+        amount -= fee;
+
         // Calculate operator commission (if any) based on remaining amount
         operatorCommission = 0;
         if (commissionRateBps > 0) {
@@ -1082,8 +1090,6 @@ contract Payments is ReentrancyGuard {
             Account storage serviceFeeRecipientAccount = accounts[token][serviceFeeRecipient];
             serviceFeeRecipientAccount.funds += operatorCommission;
         }
-
-        return (netPayeeAmount, operatorCommission);
     }
 
     function processOneTimePayment(
@@ -1103,7 +1109,7 @@ contract Payments is ReentrancyGuard {
             payer.funds -= oneTimePayment;
 
             // Calculate fees, pay operator commission and track platform fees
-            (uint256 netPayeeAmount, uint256 operatorCommission) =
+            (uint256 netPayeeAmount, uint256 operatorCommission, uint256 networkFee) =
                 calculateAndPayFees(oneTimePayment, rail.token, rail.serviceFeeRecipient, rail.commissionRateBps);
 
             // Credit payee (net amount after fees)
@@ -1114,11 +1120,11 @@ contract Payments is ReentrancyGuard {
     }
 
     /// @notice Settles payments for a terminated rail without validation. This may only be called by the payee and after the terminated rail's max settlement epoch has passed. It's an escape-hatch to unblock payments in an otherwise stuck rail (e.g., due to a buggy validator contract) and it always pays in full.
-    /// @notice The caller must include NETWORK_FEE amount of native token as a fee.
     /// @param railId The ID of the rail to settle.
     /// @return totalSettledAmount The total amount settled and transferred.
     /// @return totalNetPayeeAmount The net amount credited to the payee after fees.
     /// @return totalOperatorCommission The commission credited to the operator.
+    /// @return totalNetworkFee The fee accrued for burning FIL. 
     /// @return finalSettledEpoch The epoch up to which settlement was actually completed.
     /// @return note Additional information about the settlement.
     function settleTerminatedRailWithoutValidation(uint256 railId)
@@ -1133,6 +1139,7 @@ contract Payments is ReentrancyGuard {
             uint256 totalSettledAmount,
             uint256 totalNetPayeeAmount,
             uint256 totalOperatorCommission,
+            uint256 totalNetworkFee,
             uint256 finalSettledEpoch,
             string memory note
         )
@@ -1147,26 +1154,13 @@ contract Payments is ReentrancyGuard {
         return settleRailInternal(railId, maxSettleEpoch, true);
     }
 
-    function burnAndRefundRest(uint256 _amount) internal {
-        require(msg.value >= _amount, Errors.InsufficientNativeTokenForBurn(_amount, msg.value));
-        // f099 burn address
-        (bool success,) = BURN_ADDRESS.call{value: _amount}("");
-        require(success, Errors.NativeTransferFailed(BURN_ADDRESS, _amount));
-
-        if (msg.value > _amount) {
-            uint256 refund = msg.value - _amount;
-            (success,) = msg.sender.call{value: refund}("");
-            require(success, Errors.NativeTransferFailed(msg.sender, refund));
-        }
-    }
-
     /// @notice Settles payments for a rail up to the specified epoch. Settlement may fail to reach the target epoch if either the client lacks the funds to pay up to the current epoch or the validator refuses to settle the entire requested range.
-    /// @notice The caller must include NETWORK_FEE amount of native token as a fee.
     /// @param railId The ID of the rail to settle.
     /// @param untilEpoch The epoch up to which to settle (must not exceed current block number).
     /// @return totalSettledAmount The total amount settled and transferred.
     /// @return totalNetPayeeAmount The net amount credited to the payee after fees.
     /// @return totalOperatorCommission The commission credited to the operator.
+    /// @return totalNetworkFee The fee accrued to burn FIL.
     /// @return finalSettledEpoch The epoch up to which settlement was actually completed.
     /// @return note Additional information about the settlement (especially from validation).
     function settleRail(uint256 railId, uint256 untilEpoch)
@@ -1180,6 +1174,7 @@ contract Payments is ReentrancyGuard {
             uint256 totalSettledAmount,
             uint256 totalNetPayeeAmount,
             uint256 totalOperatorCommission,
+            uint256 totalNetworkFee,
             uint256 finalSettledEpoch,
             string memory note
         )
@@ -1193,13 +1188,11 @@ contract Payments is ReentrancyGuard {
             uint256 totalSettledAmount,
             uint256 totalNetPayeeAmount,
             uint256 totalOperatorCommission,
+            uint256 totalNetworkFee,
             uint256 finalSettledEpoch,
             string memory note
         )
     {
-        if (NETWORK_FEE > 0) {
-            burnAndRefundRest(NETWORK_FEE);
-        }
         require(untilEpoch <= block.number, Errors.CannotSettleFutureEpochs(railId, untilEpoch, block.number));
 
         Rail storage rail = rails[railId];
@@ -1208,7 +1201,7 @@ contract Payments is ReentrancyGuard {
         // Handle terminated and fully settled rails that are still not finalised
         if (isRailTerminated(rail, railId) && rail.settledUpTo >= rail.endEpoch) {
             finalizeTerminatedRail(railId, rail, payer);
-            return (0, 0, 0, rail.settledUpTo, "rail fully settled and finalized");
+            return (0, 0, 0, 0, rail.settledUpTo, "rail fully settled and finalized");
         }
 
         // Calculate the maximum settlement epoch based on account lockup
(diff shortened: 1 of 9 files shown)


src/Payments.sol
```diff
@@ -1754,6 +1783,39 @@ contract Payments is ReentrancyGuard {
...
+        auction.startTime = uint168(block.timestamp);
+
+        fees.funds = available - requested;
+
+        (bool success,) = BURN_ADDRESS.call{value: msg.value}("");
+        require(success, Errors.NativeTransferFailed(BURN_ADDRESS, msg.value));
+
+        {
+            uint256 balanceBefore = token.balanceOf(address(this));
+            token.safeTransfer(recipient, requested);
```
wjmelements, line 1807: I had the idea to do an internal balance transfer instead, but we would not want these market makers to have to perform settlement check to withdraw.

test/DepositWithAuthorizationAndOperatorApproval.t.sol
```diff
@@ -98,7 +98,7 @@ contract DepositWithAuthorization is Test, BaseTestHelper {
         );
 
         // Corrupt r
-        r = bytes32(uint256(r) ^ 1);
+        r = bytes32(uint256(r) ^ 3);
```
wjmelements, line 101: https://github.com/FilOzone/filecoin-pay/pull/229/files#r2356494903

Changing the private key made this signature valid. This change made it invalid again.

src/Payments.sol
```diff
@@ -1066,8 +1083,24 @@ contract Payments is ReentrancyGuard {
...
-        returns (uint256 netPayeeAmount, uint256 operatorCommission)
+        returns (uint256 netPayeeAmount, uint256 operatorCommission, uint256 fee)
     {
+        // ceil()
+        fee = (amount * NETWORK_FEE_NUMERATOR + (NETWORK_FEE_DENOMINATOR - 1)) / NETWORK_FEE_DENOMINATOR;
+        if (token == NATIVE_TOKEN) {
+            (bool success,) = BURN_ADDRESS.call{value: fee}("");
+            require(success, Errors.NativeTransferFailed(BURN_ADDRESS, msg.value));
+        } else {
+            accounts[token][address(this)].funds += fee;
```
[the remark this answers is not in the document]
wjmelements: related https://github.com/FilOzone/filecoin-pay/issues/218

src/Payments.sol
```diff
@@ -35,13 +37,17 @@ interface IValidator {
 
 // @title Payments contract.
 contract Payments is ReentrancyGuard {
+    using Dutch for uint256;
     using SafeERC20 for IERC20;
     using RateChangeQueue for RateChangeQueue.Queue;
 
     // Maximum commission rate in basis points (100% = 10000 BPS)
     uint256 public constant COMMISSION_MAX_BPS = 10000;
+    uint88 private constant AUCTION_START_PRICE = 0.083 ether; // 0.083 FIL
```
rvagg, line 46: I'm sure I missed this in a doc somewhere but what is this based on as a start? I don't think we special-case USDFC in here anywhere so this is a generic start price for any ERC20, so how do we get to 0.083?

wjmelements: it's in FIL. research picked this price. Maybe should be 4x higher to account for 1 week of decay

src/Payments.sol
```diff
@@ -1754,6 +1784,38 @@ contract Payments is ReentrancyGuard {
 
         return (fundedUntilEpoch, currentFunds, availableFunds, currentLockupRate);
     }
+
+    function burnFILForFees(IERC20 token, address recipient, uint256 requested) external payable {
```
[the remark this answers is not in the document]
wjmelements: This is the auction accept function that converts fees into fil and burns the fil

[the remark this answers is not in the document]
rvagg: oh right sorry, this is the public primary auction function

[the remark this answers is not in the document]
wjmelements: I agree this function is missing doc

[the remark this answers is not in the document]
wjmelements: 93c0709

src/Payments.sol
```diff
@@ -1754,6 +1782,44 @@ contract Payments is ReentrancyGuard {
...
+
+        {
+            uint256 balanceBefore = token.balanceOf(address(this));
+            token.safeTransfer(recipient, requested);
+            uint256 balanceAfter = token.balanceOf(address(this));
+
+            uint256 balanceChange = balanceBefore - balanceAfter;
+            // handle fee-on-transfer and hidden-denominator tokens
+            if (balanceChange < requested) {
+                fees.funds += requested - balanceChange;
```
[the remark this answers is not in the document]
wjmelements: I'll probably just remove this case.

src/Dutch.sol
```diff
@@ -0,0 +1,28 @@
...
+    function decay(uint256 startPrice, uint256 elapsed) internal pure returns (uint256 price) {
+        if (elapsed > MAX_DECAY) {
+            return 0;
+        }
+        UD60x18 coefficient = UD60x18.wrap(startPrice);
+        UD60x18 decayFactor = UD60x18.wrap(elapsed * uUNIT / HALVING_INTERVAL).exp2();
+
+        return coefficient.div(decayFactor).unwrap();
+    }
+}
```
lucaniz, line 28: Seems correct to me, comparing with the formula in https://www.notion.so/filecoindev/Liquidation-Price-Modeling-26cdc41950c180dca05de4c991d56180?source=copy_link#270dc41950c18024b568ebc9341b2d6d 

Basically, we put RESET_FACTOR = 4, halving interval 3.5 days which means startPrice is divided by 4 after 2 halving interval (i.e. 1 week), which matches the formula in the notion with M = 4. (as here we have price = startPrice / 2^(elapsed / HALVING_INTERVAL) ) cc @wjmelements

src/Payments.sol
```diff
@@ -35,15 +37,16 @@ interface IValidator {
...
 // @title Payments contract.
 contract Payments is ReentrancyGuard {
+    using Dutch for uint256;
     using SafeERC20 for IERC20;
     using RateChangeQueue for RateChangeQueue.Queue;
 
     // Maximum commission rate in basis points (100% = 10000 BPS)
     uint256 public constant COMMISSION_MAX_BPS = 10000;
-    uint256 private constant AUCTION_START_PRICE = .083 ether; // 0.083 FIL
+    uint88 private constant AUCTION_START_PRICE = 10.4 ether; // 10.4 FIL
```
lucaniz, line 46: where did we pick the 10.4 FIL value? if we consider 300TiB network size and PDP + Payment fee (and payment fee being implemented with the 0.5% of the volume, with the minimum fee value), then it should be something like 7.8 i think

lucaniz: Great. This value assume a network size of ~10 TiB, we need to decide what is the starting point we want to set up (whether it is the target network size for having 50USDFC transacted every day, i.e. 300 TiB, or a smaller initial size of 10TiB)

wjmelements: Increasing this to 7.83×4=31.32 FIL per our discussion. It's slightly safer to start high.

src/Payments.sol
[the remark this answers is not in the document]
wjmelements: cc5cfed

src/Payments.sol
Stebalien: IMO (as a now outsider) we really should do this as an internal transfer. Market makers are already calling smart contract methods, they can handle a withdraw call and it makes it easier to reason about the security here.

src/Payments.sol
```diff
@@ -1754,6 +1783,43 @@ contract Payments is ReentrancyGuard {
...
+     * @param requested Exact amount of fees transferred
+     */
+    function burnFILForFees(IERC20 token, address recipient, uint256 requested) external payable {
+        Account storage fees = accounts[token][address(this)];
+        uint256 available = fees.funds;
+        require(available >= requested, Errors.WithdrawAmountExceedsAccumulatedFees(token, available, requested));
+
+        AuctionInfo storage auction = auctionInfo[token];
+        uint256 auctionPrice = uint256(auction.startPrice).decay(block.timestamp - auction.startTime);
+        require(msg.value >= auctionPrice, Errors.InsufficientNativeTokenForBurn(msg.value, auctionPrice));
```
Stebalien, line 1800: `requested * auctionPrice`?

src/Payments.sol
```diff
@@ -1754,6 +1783,43 @@ contract Payments is ReentrancyGuard {
...
+
+        auction.startPrice = uint88(auctionPrice * Dutch.RESET_FACTOR);
+        auction.startTime = uint168(block.timestamp);
+
+        fees.funds = available - requested;
+
+        (bool success,) = BURN_ADDRESS.call{value: msg.value}("");
```
Stebalien, line 1807: I'd transfer with method 0, for gas and safety (assuming we're sending to f099).

src/Payments.sol
```diff
@@ -1754,6 +1783,43 @@ contract Payments is ReentrancyGuard {
...
+        require(success, Errors.NativeTransferFailed(BURN_ADDRESS, msg.value));
+
+        {
+            uint256 balanceBefore = token.balanceOf(address(this));
+            token.safeTransfer(recipient, requested);
+            uint256 balanceAfter = token.balanceOf(address(this));
+
+            uint256 balanceChange = balanceBefore - balanceAfter;
+            // handle fee-on-transfer and hidden-denominator tokens
+            // re-entry via erc20 transfer can burn unrequested fees, which is fine
```
Stebalien, line 1817: I'm not sure what this comment is trying to say.

In general, one would _usually_ update the funds _first_, then transfer as the very last step to avoid any reentrancy issues. If the transfer fails, you'd simply revert everything.

src/Payments.sol
wjmelements: the price is for the full balance, not per-wei

Stebalien: But I'm buying `requested` which may be less than the full balance.

src/Payments.sol
wjmelements: ERC20 tokens can do all sorts of weird things. What this comment is considering is if that ERC20 transfer() gave execution back to a malicious party who can re-enter Payments: what can they do with this power? They can settleRail; they can deposit or withdraw.

> one would usually update the funds first, then transfer as the very last step to avoid any reentrancy issues.

That is what we do, and then we do balanceOf and adjust the auction balance again, but only downward, in case the transfer sent out more tokens than we expected (example of this in `test/mocks/ExtraFeeToken.sol`, tested in `test/BurnExtraFeeToken.t.sol`).

src/Payments.sol
wjmelements: what does this mean?

src/Payments.sol
```diff
@@ -1754,6 +1783,47 @@ contract Payments is ReentrancyGuard {
...
         return (fundedUntilEpoch, currentFunds, availableFunds, currentLockupRate);
     }
+
+    /**
+     * @notice Burn FIL to buy the network fees
+     * @param token Which kind of fees to buy
+     * @param recipient Receives the purchased fees
+     * @param requested Exact amount of fees transferred
+     */
+    function burnFILForFees(IERC20 token, address recipient, uint256 requested) external payable {
```
Stebalien, line 1793: It's very hard to use this method to reliably buy a token for a specific price:

1. It'll fail if `requested` tokens aren't available even if there are SOME tokens available. E.g., if many people are racing to participate in the auction at the same time. Ideally, I'd be able to buy (a) at most X tokens (b) at at most Y price.
2. If the price goes down and I send too much FIL, that FIL will get burnt anyways. Really, I'm _guaranteed_ to lose some FIL because it'll always take some time for my message to be included. Ideally, this method would pull the exact amount FIL required from my account within the payment system (this would also ensure that there's exactly ONE way to move funds into and out of the payment system).

src/Payments.sol
wjmelements: `withdraw` does `settleAccountLockupBeforeAndAfter` which does 2x `performSettlementCheck` etc. I don't think these make sense when they probably are neither a payer nor a payee. They probably will want to withdraw in the same transaction and we should make that cheap for them since the gas cost of completing this auction effectively reduces the FIL burn.

Stebalien: 1. Those calls will cost nothing (or almost nothing) if there are no rails, etc.
2. I expect anyone buying tokens here will be making multiple purchases over time.

You can implement this however you want, but it'll be a lot easier to reason about if there's exactly one way to deposit and withdraw funds.

If you do keep this as-is, I'd apply the global re-entrancy guard just to be safe and to avoid having to reason about any of this.

src/Payments.sol
Stebalien: Ok, none of this code is making any sense to me now. Are you saying:

1. The recorded price is the price for the entire bucket.
2. I can only buy out the entire bucket.
3. I always have to pay for the entire bucket.
4. I have to specify how much is in the bucket. If I over shoot, my message fails. If I under shoot, I over pay.

If that's what's happening here, I have no idea why you'd do it this way and it'll DEFINITELY surprise users.

src/Payments.sol
Stebalien: Please look at this more carefully. This code is _double accounting_ for deducted funds (funds that were already accounted for on withdrawal).

But as I no longer work here, I can't really spend any more time on this.

wjmelements: @zenground0 might have an opinion on this since it's similar to #165. Maybe we also want to do the wrapping `balanceOf` for withdrawal. I should probably also make this method `nonReentrant` to hedge against my own hubris.

wjmelements: 708662a

src/Payments.sol
ZenGround0: `0xff00000000000000000000000000000000000063` is f099.

src/Payments.sol
ZenGround0: >  I allow them to take less than the full amount to handle weird token balance behaviors and also recipients like UniswapV4 that might want less than the full amount.

For simplicity's sake I suggest we don't handle these cases and force the buyer to simply buy the whole amount.  Especially since
> I expect requested amount to not be much less than the balance.

src/Payments.sol
wjmelements: I really think it's unnecessary and adds a lot of overhead that would reduce the FIL burn.

> very clean

I suspect not. We don't have other code for internal transfers so there's also complexity there. An internal transfer would probably need to do the same lockup operations as deposit.

>  writes a slot, emits an event

This happens twice: once for the native token (FIL), and once for the purchased token.

> a handful of slot reads

This adds deposit native and withdraw token to the operation. Here are the unnecessary reads that adds, assuming they have lockupRate=0:
* 4: funds and currentLockup for token and native 
* 1: token lockupFullySettled
* 2: native and token lockupLastSettledAt

src/Payments.sol
ZenGround0: If I am underestimating how painful this makes it for people to participate then please elaborate on the weird token balance behaviors or why uniswap would want to buy only some.

src/Payments.sol
Stebalien: @ZenGround0 yes. TBH, it's probably fine either way.

src/Payments.sol
wjmelements: I think you're overeager to give favors to the bots, who are not our users. If they are willing to take less tokens we should keep them for a later auction. If they are willing to burn more FIL than the current time dictates we should let them. At the time they signed the transaction they were willing to burn that much FIL for that many tokens. The occasional windfall should be in our favor, not theirs.

> why uniswap would want to buy only some

In UniswapV4 the balance changes all have to zero out or it will revert.
In UniswapV2 the balances overflow at a fairly low threshold (uint96).

src/Payments.sol
ZenGround0: That's not an issue as precompiles work fine in tests if you set things up: https://github.com/FilOzone/pdp/blob/main/src/PDPVerifier.sol#L615 

I think its probably not a real safety issue (though please correct me if I'm wrong)

But @wjmelements since you are very motivated to reduce gas costs it is worth looking into the difference.

src/Payments.sol
ZenGround0: Oh so the point is to have an interface compatible with tokens like this ?

wjmelements: > Oh so the point is to have an interface compatible with tokens like this ?

We'd like to support as many tokens as we can. I did originally transfer-out `minOut` tokens but then I remembered the fee-on-transfer problem in #165.

wjmelements: > There are tokens where transfer decreases balance by more than the specified amount

The hidden denominator token is the most common way it manifests. These tokens are sometimes called rebase tokens. Their `balanceOf` and `transfer` balances are actually a numerator and a denominator. The decnominator can change any time and the numerator is your true balance, but you can only specify and query the divided amount. Very annoying to deal with. Quite popular around 2021-2022. A recent examples is USDM.

rvagg: I think the argument that makes this not unreasonable is that this is going to be entirely bot controlled, with parameters for acceptable levels of slippage.

Also, if a message lands on this to extract the amount that is believed to be in there but there's now more in there because there was some previous settlement method, then if we made them take the whole pot then there's a nice win there and some MEV potential .. I think (within the bounds of what MEV you can really pull off with Filecoin's message selection & execution system)? If you have to say how much of it you want and for how much then you can hit the exact amount you want/need and not be in for surprises if you win. If we force the requested amount to `==` the amount available, then we are in the message ordering problem again where you're just taking guesses of what it's going to be. So maybe the best world here is the one that's implemented - you say how much you want and how much you're willing to spend, if you undershoot for the pot then that's just how it is, but rational actors are going to try and get it spot on to maximise value.

Then again, I could just be making stuff up here. I've been trying to think this logic through but it's not my forte, so I'm not used to reasoning about this.

src/Payments.sol
rvagg: I couldn't help exploring this a little further to understand the suggestion: https://github.com/FilOzone/filecoin-pay/pull/234

As per comment in there, I'm not sure it's a better option here. Maybe it ends up being better in gas (I kind of doubt it), but how much is gas going to factor into settlement price for the auction so do we care?

wjmelements:
> * Why is `HALVING_INTERVAL` set to 3.5 days instead of shorter/longer?
> * Why is the `RESET_FACTOR` 4?

log2(4) = 2, 2 * 3.5 days = 7 days. See the description:
```
We can expect approximately one auction per token per week.
```
We target 7 days because that matches the tempo of gas prices in congested networks. If you have a strong reason to pick a different interval that would easily outweigh this reasoning.

As for the reset multiple, there is a tradeoff in price precision vs regularity. If the multiple were larger there would be less variance in auction duration but also less price precision. For example if the fees were accelerating, a higher multiple would settle the auction closer to the (1 week) interval target than a lower multiple, but the price change per epoch would also be larger, and this would reduce the FIL burn. I'm not too opinionated on the multiple; I originally suggested between 4 and 16. Powers of two are slightly convenient for the math. For now we've agreed on 4 but it would not be hard to tweak these before launch.

> What happens if these assumptions prove wrong?

It is possible that the work week would change, eg to 10 days. The consequence would not be significant. Other situations like rapid changes in the fee rate are handled pretty well.

> From what I've learned, many DeFi protocols evolve parameters over time without requiring full upgrades or parallel deployments. It might be worth considering a mechanism for adjusting these safely.

We choose to let the price float instead of the duration because we want these auctions to settle at regular intervals so that we can price the fees into FIL, which is important for some ideas downstream of this change like datacap allocation. If the auction is too irregular there could be a delay in such allocation because the pricing would be delayed.

> require payment rails to be backed by DEX pools with sufficient liquidity into FIL

This would require enshrining a particular DEX interface and those are evolving rapidly. More critically though, liquidity is very easy to fake. There is no solution to this problem. Any DEX test or check we could do is only valid at the moment of execution and we cannot extrapolate beyond that, even within the same transaction.

> Allowing “anything goes” ERC20 payment rails could introduce UX and security risks.
> Malicious ERC20s: what if balanceOf lies or balances shift unexpectedly? Could payouts fail?
> Honeypot / blacklist / pausable tokens: can users deposit but later fail to withdraw?
> Transfer reverts: do failed ERC20 transfers risk bricking the rail?


We have to allow for this. ERC20 being an interface, there is no way to check if a token is valid, short of formal verification of their code, and proxy tokens that we want to support, such as USDFC, would fail such formal verification. Really the only alternative is curation via some kind of governance. But we want to limit our governance to the maximum possible extent, so instead we let users use whatever tokens they want. They will do their own token curation. If they use a token that cannot withdraw, that is their choice.

> My intuition is that rails are non-contagious (a bad rail doesn’t poison others)

If a token is bad it won't break any of the flow for other tokens.

wjmelements:
> could it make sense for each payment rail to be its own contract, deployed at its own address?

This could be fine but it's out of scope for this PR. The use cases we have in mind for now do benefit from a shared payments contract, but they could instead consult a factory registry. It would add a level of indirection, but it would work.

> A library of payment rails could be made available to deploy through this contract, and the deployer of a rail could be responsible for governing/upgrading it

What you're describing is similar to a package manager, where end users can pull the latest version or even rollback. The best kinds of upgrades would also simplify the storage scheme, but such a migration can be unsafe for blockchain applications, partly because the cost can exceed the max gas. It is also not generally possible to know if some migrations have completed because most mappings are not iterable. So the scope of the upgrade would be constrained by prior versions.

My bigger gripe with upgradeable Payments is that any kind of governance could result in a bad version release (accidental or malicious). Decentralizing the governance as you suggest does reduce the risk to end users because there is a second level of defense. But even better would be for such breakage to be possible. Safe immutable code would remain safe and does not have custodial risk. Users can move to newer versions at their own risk. This is the most consensual model, most similar to a real package manager.

wjmelements:
> I think a UUPSUpgradeable contract here would be a simple start

We had it before and I removed it after a lot of discussion and research (led by @ZenGround0)  
https://github.com/FilOzone/filecoin-pay/pull/196
Other parts of the stack are UUPS but we think Payments is especially important to be non-custodial.

> Challenge here (imo) is less technical, more governance. Agree it adds surface area of attack vulnerability.

Yes, and we want to minimize that.

> that putting this system into production and getting real feedback from users

We are collecting feedback from developers, users, and storage providers on the whole stack during the development process. For this particular change I have some test cases to make sure it seems reasonable, especially in extreme cases, but we may tweak these parameters several times before launch.

src/Payments.sol
wjmelements: I tried measuring gas with my `forge script` method but I hit this foundry bug. https://github.com/foundry-rs/foundry/issues/11731

So let's make this a separate PR. I will create an issue to track it.

src/Payments.sol
```diff
@@ -6,10 +6,17 @@ import "@openzeppelin/contracts/token/ERC20/extensions/IERC20Permit.sol";
...
+
+import {Dutch} from "./Dutch.sol";
 import "./Errors.sol";
 import "./RateChangeQueue.sol";
 import "./interfaces/IERC3009.sol";
 
+uint88 constant UINT88_MAX = 0xffffffffffffffffffffff;
+
+uint88 constant MAX_AUCTION_START_PRICE = UINT88_MAX; // 309485009.821345068724781055 FIL
+uint88 constant FIRST_AUCTION_START_PRICE = 31.32 ether; // 31.32 FIL
```
wjmelements, line 18: @lucaniz @irenegia 
This is the constant most likely to be wrong. I think it's too large. When we were picking this number we were reasoning about a recurring reverse-dutch auction, and this number reflects our goals for a mature network. But since then we have changed the design to double-dutch, so auction price fluctuates algorithmically now, and this constant only determines the start price of the first auction.

I foresee that the first usage of a token might have less volume. One exception would be if an existing operator (eg FWSS) switched their preferred token and were able to get their users onto new rails. I don't think such a migration would be fast because operator approval is per-token. So I don't think this first value needs to be too large.

If this number is too large then the first burn for that token would be delayed. If it is too small then there might be less burn for the first auction. I think, for the first auction, a delay in the burn is worse than price imprecision because it can impact the user experience. I also think we are getting decent precision from the week duration.

For example, if the actual starting fee rate was worth 0.02 FIL per week, we can [expect](https://www.wolframalpha.com/input?i=Solve+for+t%3A+31.32+*+4%5E%28-t%29+%3D+.02+*+t) the first auction to finish after about 30 days. Maybe this is fine. The productlog nature of the auction duration limits the impact of setting the value too large considerably.

Next let's consider the possibility that this starting price is actually too small. The 75% decay per week is 0.006876% per epoch, and there are 20160 epochs per week. If the FIL value of the fees were 20160 FIL per week, the first auction [should](https://www.wolframalpha.com/input?i=Solve+for+t%3A+31.32+*+4%5E%28-t%2F20160%29+%3D+t) finish around the 32nd epoch, also where we would expect it to finish if the first auction closed in the same block that it began. The price precision here only costs us about 2.2%, and the worst rate that would also finish the auction in 32 blocks, imprecision would reduce the burn 1/32 = 3.125% from the ideal. So even in this extreme scenario with antagonistic numbers we can see this number is not too low.

Perhaps this imprecision analysis is the correct way to determine the first auction start price. We should pick the largest starting fee rate (eg 300 FIL/week) and price imprecision (eg 1%) we can tolerate the first auction imprecision. So for example for 1% imprecision we would calculate for 99 epochs (because 99/(99+1) = 99%); at 300 FIL per week this is `300 * (99/20160) = 1.4732142857142856` FIL. Similarly if we only wanted to tolerate 0.1% imprecision for the first auction would calculate for 999 epochs; at 300 FIL per week this is `14.866071428571427` FIL. 

Let me know what you think!

rvagg (approved):

src/Payments.sol
lucaniz: So the possible (potentially long) delay of the first auction is something I highlighted on slack as well and I had different feedback on whether this was an issue or not (for instance, iirc @Kubuxu said in his opinion it was not a big deal). 

Given this issue is related to the very first auction only, I think we can put the number small enough, maybe assuming an initial target size. 
I am not fully following your numbers, will try to get back on the details

PR #233, opened 2025-09-18
Withdrawal balanceOf Check

wjmelements:
Reviewer @zenground0
For the same reason as deposit, we need to check balanceOf before and after withdrawal to see if the balance was reduced by more than expected.

It should not be possible that one user depositing should allow another to withdraw. This implies system debt.
#### Changes
* check balanceOf before and after withdrawal

### DIFF
--- a/test/WithdrawExtraFeeToken.t.sol
+++ b/test/WithdrawExtraFeeToken.t.sol
@@ -72,12 +72,19 @@ contract WithdrawExtraFeeTokenTest is Test {
         uint256 transferFee = 10 ** 18;
         ExtraFeeToken feeToken = new ExtraFeeToken(transferFee);
         address user1 = vm.addr(0x1111);
+        address user2 = vm.addr(0x1112);
         feeToken.mint(user1, 10 ** 24);
+        feeToken.mint(user2, 10 ** 24);
 
         vm.prank(user1);
         feeToken.approve(address(payments), 10 ** 24);
         vm.prank(user1);
         payments.deposit(feeToken, user1, 10 ** 24 - transferFee);
+
+        vm.prank(user2);
+        feeToken.approve(address(payments), 10 ** 24);
+        vm.prank(user2);
+        payments.deposit(feeToken, user2, 10 ** 24 - transferFee);
 
         (uint256 deposit,,,) = payments.accounts(feeToken, user1);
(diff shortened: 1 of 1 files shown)


wjmelements:
```
All modified and coverable lines are covered by tests.

❌ Your project check has failed because the head coverage (75.64%) is below the target coverage (85.00%). You can increase the head coverage or adjust the [target](https://docs.codecov.com/docs/commit-status#target) coverage.
```
that's annoying

src/Payments.sol
```diff
@@ -801,7 +801,13 @@ contract Payments is ReentrancyGuard {
             (bool success,) = payable(to).call{value: amount}("");
             require(success, Errors.NativeTransferFailed(to, amount));
         } else {
+            uint256 balanceBefore = token.balanceOf(address(this));
             token.safeTransfer(to, amount);
+            uint256 balanceAfter = token.balanceOf(address(this));
+            uint256 balanceChange = balanceBefore - balanceAfter;
+            if (balanceChange > amount) {
+                account.funds -= balanceChange - amount;
```
wjmelements, line 809: also need to check against `available` here

wjmelements: 7e0e4aa

test/WithdrawExtraFeeToken.t.sol
```diff
@@ -0,0 +1,68 @@
...
+        payments.deposit(feeToken, user2, 10 ** 23);
+        (deposit,,,) = payments.accounts(feeToken, user2);
+        assertEq(deposit, 10 ** 23);
+
+        assertEq(feeToken.balanceOf(address(payments)), 2 * 10 ** 23);
+
+        // the other user's deposit should not allow the withdrawal
+        vm.prank(user1);
+        vm.expectRevert();
+        payments.withdraw(feeToken, 10 ** 23);
```
wjmelements, line 49: this is the test that fails without the fix

wjmelements:
I spent some time investigating and I disagree with the coverage report. Both the `InsufficientUnlockedFunds` cases have test coverage, but it says neither of them do.

wjmelements:
Similar to #165, we don't expect this from USDFC or axlUSDC, but other tokens might break this balance assumption for whatever reason. We would like to support such tokens to the extent possible while protecting our users' deposits, but even this cannot work for every conceivable scheme.

PR #234, opened 2025-09-19
feat: burn via precompile

rvagg:
I got nerdsniped by the discussion in https://github.com/FilOzone/filecoin-pay/pull/229#discussion_r2360820122 and couldn't help exploring this a little (with some help from AI). It seems that mocks work OK for testing. I haven't tried deploying this and watch it working so there's a bit of guessing going on here.

But overall, unless there really is a compelling safety or gas reason for this I don't think this is a great idea—we'll have people looking at this code and learning from it, do we want to express this as the right way to transfer funds in contracts?

PR just for interest's sake, will discard this.


src/Payments.sol
```diff
@@ -1823,6 +1823,31 @@ contract Payments is ReentrancyGuard {
...
     }
+
+    /**
+     * @dev Burns FIL using the FVM precompile to call the burn actor (ID 99)
+     * @param amount The amount of FIL to burn
+     */
+    function burnViaPrecompile(uint256 amount) private {
+        // Using simpler encoding to reduce stack usage
+        bytes memory empty = "";
+        bytes memory data = abi.encode(
```
wjmelements, line 1834: should this be encodePacked?

src/Payments.sol
```diff
@@ -1823,6 +1823,31 @@ contract Payments is ReentrancyGuard {
...
+            uint64(0),     // method 0
+            amount,        // value
+            uint64(0),     // flags
+            uint64(0),     // codec
+            empty,         // params
```
wjmelements, line 1839: perhaps it really is abi encoded

src/Payments.sol
```diff
@@ -1823,6 +1823,31 @@ contract Payments is ReentrancyGuard {
...
+    /**
+     * @dev Burns FIL using the FVM precompile to call the burn actor (ID 99)
+     * @param amount The amount of FIL to burn
+     */
+    function burnViaPrecompile(uint256 amount) private {
+        // Using simpler encoding to reduce stack usage
+        bytes memory empty = "";
+        bytes memory data = abi.encode(
+            uint64(0),     // method 0
+            amount,        // value
```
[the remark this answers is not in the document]
rvagg: it's a 32: https://github.com/filecoin-project/builtin-actors/blob/master/actors/evm/src/interpreter/precompiles/fvm.rs#L138

wjmelements:
I think this fails in eth_estimateGas so I am looking into why.

wjmelements:
> I think this fails in eth_estimateGas so I am looking into why.

It's actually failing in a different way. I don't think vm.etch works in forge scripts.

wjmelements:
> > I think this fails in eth_estimateGas so I am looking into why.
> 
> It's actually failing in a different way. I don't think vm.etch works in forge scripts.

Filed an issue: https://github.com/foundry-rs/foundry/issues/11731

src/Payments.sol
```diff
@@ -1804,15 +1805,39 @@ contract Payments is ReentrancyGuard {
...
+     */
+    function burnViaPrecompile(uint256 amount) private {
+        // Using simpler encoding to reduce stack usage
+        bytes memory empty = "";
+        bytes memory data = abi.encode(
+            uint64(0), // method 0
+            amount, // value
+            uint64(0), // flags
+            uint64(0), // codec
+            empty, // params
```
wjmelements, line 1829: Since this is an empty buffer, there is an encoding trick you can do where you point this to another zero you already have in the calldata. For example, because your `method` is 0, you can put `0` for the `params`. This encoding uses slightly less data.

wjmelements: 8e09330

wjmelements:
Based on https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0057.md I expect payment via method 0 will be cheaper.

wjmelements:
I think the merge conflicts are because of the squash-merge I did. Planning to rebase this branch to fix it.

wjmelements:
I have a preliminary measurement showing that this saves 5,215,069 gas, but that seems too high. Gonna validate the result some more.

wjmelements:
> I have a preliminary measurement showing that this saves 5,215,069 gas, but that seems too high. Gonna validate the result some more.

The bulk of this difference was because I was sending the amount to a non-existent t410. The difference was smaller when sending to an existing account and even smaller when sending to the burn address. The actual ~460k savings is less than the 500k+75k I expected from [this doc](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0057.md) but still quite good.

[burn via precompile](https://calibration.filfox.info/en/message/0xc59fee35a3e1baa68b1b1b2b22fbfe1a8d156721162e07a1c115613aadd0d7d9) used 2,858,069 / 3,567,711

[burn via evm CALL](https://calibration.filfox.info/en/message/0xfdc18e51a4b1e6f6e00e85c4c563d9620a2c34698197b996ed216eab97a9ee06) used 3,318,405 / 4,143,131

[first evm CALL to an account](https://calibration.filfox.info/en/message/0xe4bd9d0c09cc3f78bde481537c04dcb3c401276d708616a7acaaaafa7cd5db4c) used 7,019,892 / 8,769,990

[second evm CALL to an account](https://calibration.filfox.info/en/message/0xeae2684967f60ce4f970b220e7c62a6bca6075228a52b02870f0268d947f9817) used 4,319,892 / 5,394,990

src/Payments.sol
```diff
@@ -1821,12 +1822,36 @@ contract Payments is ReentrancyGuard {
...
+            mstore(fmp, 0) // method 0
+            mstore(add(32, fmp), amount) // value
+            mstore(add(64, fmp), 0) // flags
+            mstore(add(96, fmp), 0) // codec
+            mstore(add(128, fmp), 0) // params
+            mstore(add(160, fmp), BURN_ACTOR_ID) // actor ID
+            ok :=
+                and(
+                    and(gt(returndatasize(), 31), eq(mload(fmp), 0)),
+                    delegatecall(gas(), CALL_ACTOR_ID_PRECOMPILE, fmp, 192, fmp, 192)
```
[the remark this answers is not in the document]
wjmelements: https://github.com/filecoin-project/fvm-solidity

wjmelements (approved):
would like a review from @ZenGround0 @rvagg or @Kubuxu

Kubuxu (approved):
It seems like quite a bit of complexity for not a lot of gain, but it looks easy to revert if we have to. SGWM.

src/Payments.sol
```diff
@@ -1821,8 +1822,7 @@ contract Payments is ReentrancyGuard {
         auction.startPrice = uint88(auctionPrice);
         auction.startTime = uint168(block.timestamp);
 
-        (bool success,) = BURN_ADDRESS.call{value: msg.value}("");
-        require(success, Errors.NativeTransferFailed(BURN_ADDRESS, msg.value));
+        require(msg.value.burn(), Errors.NativeTransferFailed(BURN_ADDRESS, msg.value));
```
Kubuxu, line 1825: Similar comment as in PDP, explicit call will be less "magical" and more readable.

PR #240, opened 2025-09-22
forge lint

wjmelements:
Reviewer @rvagg
#### Changes
* add `forge lint` to `lint` workflow
* TODO fix lint

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -18,6 +18,10 @@ jobs:
         with:
           version: v1.2.3
 
-      - name: Lint
+      - name: fmt
         run: |
           forge fmt --check
+
+      - name: Lint
+        run: |
+          forge lint 2>&1 | read -t 0


.github/workflows/lint.yml
```diff
@@ -18,6 +18,10 @@ jobs:
...
           version: v1.2.3
 
-      - name: Lint
+      - name: fmt
         run: |
           forge fmt --check
+
+      - name: Lint
+        run: |
+          forge lint 2>&1 | read -t 0
```
wjmelements, line 27: this is the wrong check. I'll try something else

.github/workflows/lint.yml
```diff
@@ -24,4 +24,4 @@ jobs:
 
       - name: Lint
         run: |
-          forge lint 2>&1 | read -t 0
+          forge lint 2>&1 | test ! -s /dev/stdin
```
wjmelements, line 27: surprisingly this doesn't work either

src/Payments.sol
```diff
@@ -181,7 +181,7 @@ contract Payments is ReentrancyGuard {
         string note;
     }
 
-    constructor() ReentrancyGuard() {
```
wjmelements, line 184: The lint error for this line was:
```
error: modifier-style base constructor call without arguments
   --> /Users/will/projects/filecoin-pay/src/Payments.sol:184:19
    |
184 |     constructor() ReentrancyGuard() {
    |                   ^^^^^^^^^^^^^^^^^
    |
```
This is because the superclass constructor is called implicitly if it has no arguments.

.github/workflows/lint.yml
```diff
@@ -16,8 +16,16 @@ jobs:
...
         run: |
           forge fmt --check
+
+      - name: build
+        run: |
+          forge build
+
+      - name: Lint
+        run: |
+          ! (forge lint 2>&1 | grep "")
```
wjmelements, line 31: Seems we can use `--deny=warnings` starting on v1.4.0 but that version is not yet released.

PR #245, opened 2025-10-07
chore: update auction start price to 0.0021

Kubuxu:
@lucaniz for the reasoning

### DIFF
--- a/src/Payments.sol
+++ b/src/Payments.sol
@@ -17,7 +17,7 @@ uint88 constant UINT88_MAX = 0xffffffffffffffffffffff;
 
 // FIL max supply cap is 2 billion
 uint88 constant MAX_AUCTION_START_PRICE = UINT88_MAX; // 309,485,009.821345068724781055 FIL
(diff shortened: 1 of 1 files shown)


lucaniz:
### **First Auction Price **

**Chosen parameter**

Starting price = 0.0021 FIL

**Rationale**

Proposal is driven by multiple factors

- Main driver is we want the very first auction to close in days and we can not accept first auction to close in weeks (input from FilOz leadership)
- The updated target for volume transacted is 50 USD/week, which is much smaller than the initial volume taken into account in the first design doc (where it was 50USDFC/daily)
- The expectation is that PDP network will stay below 1 TiB for the first weeks and imprecision analysis covers us up until a network size of ~200 TiB (i.e. if we had 200 TiB network, an auction starting at 0.0021 FIL will take at least 50 min to close).

**Note**

This value is for the very first round only, as the auction "self adjust” over time considering the fee volumes of the former round.


More infos [here](https://www.notion.so/filecoindev/Auction-Parameters-285dc41950c18039b84cf12bd5b2db0d#285dc41950c1807db2e9dd98f60fc586)

PR #248, opened 2025-10-16
perf: Apportion fees after settling segments instead of per segment

### LINKED ISSUE #228: perf: Apportion fees after settling segments instead of per segment.
Currently we calculate and award operator commission fees within each segment. This can be highly redundant and also complicates the bookkeeping in `_settleWithRateChanges`. If the commission doesn't change over time then the fees should be calculated at the end.

Chaitu-Tatipamula:
## Summary

This PR implements the performance optimization requested in #228 to reduce redundant fee calculations during rail settlement.

## Problem

Currently, operator commission fees are calculated and awarded within each settlement segment. This approach:
- Creates redundant fee calculations when processing multiple segments
- Complicates bookkeeping in `_settleWithRateChanges` 
- Applies network fee ceiling multiple times instead of once

## Solution

**Key Changes:**
1. **Introduced `_settleSegmentGross` helper** - Handles settlement logic (debit payer, update lockup, set epochs) without fee calculation
2. **Refactored single-segment settlement** - Use gross settlement + single fee application at the end
3. **Refactored rate-changes settlement** - Aggregate gross amounts across segments, then apply fees once at completion/early exit
4. **Maintained backward compatibility** - Existing `_settleSegment` interface unchanged

**Benefits:**
- ✅ Reduces redundant `calculateAndPayFees` calls from N segments to 1
- ✅ Simplifies bookkeeping in `_settleWithRateChanges` 
- ✅ Applies network fee ceiling once instead of per segment
- ✅ Maintains exact same fee amounts and behavior
- ✅ No breaking changes to public interfaces

## Implementation Details

The change splits settlement into two phases:
1. **Gross Settlement**: Debit payer funds, update lockup, advance epochs
2. **Fee Application**: Calculate and credit fees once on total gross amount

Fixes #228

### DIFF
--- a/src/FilecoinPayV1.sol
+++ b/src/FilecoinPayV1.sol
@@ -1247,7 +1247,7 @@ contract FilecoinPayV1 is ReentrancyGuard {
 
         // Process settlement depending on whether rate changes exist
         if (rail.rateChangeQueue.isEmpty()) {
-            (totalSettledAmount,, note) =
+            (totalSettledAmount, note) =
                 _settleSegmentGross(railId, startEpoch, maxSettlementEpoch, rail.paymentRate, skipValidation);
 
             require(
@@ -1351,7 +1351,7 @@ contract FilecoinPayV1 is ReentrancyGuard {
             }
 
             // Settle the current segment with gross amounts (no fee calculation)
-            (uint256 segmentGrossSettled,, string memory validationNote) =
+            (uint256 segmentGrossSettled, string memory validationNote) =
                 _settleSegmentGross(railId, processedEpoch, segmentEndBoundary, segmentRate, skipValidation);
 
             // If validator returned no progress, exit early
@@ -1412,19 +1412,19 @@ contract FilecoinPayV1 is ReentrancyGuard {
         uint256 epochEnd,
         uint256 rate,
         bool skipValidation
-    ) internal returns (uint256 grossSettledAmount, uint256 settledUntilEpoch, string memory note) {
+    ) internal returns (uint256 grossSettledAmount, string memory note) {
         Rail storage rail = rails[railId];
         Account storage payer = accounts[rail.token][rail.from];
 
         if (rate == 0) {
             rail.settledUpTo = epochEnd;
-            return (0, epochEnd, "Zero rate payment rail");
+            return (0, "Zero rate payment rail");
         }
 
         // Calculate the default settlement values (without validation)
         uint256 duration = epochEnd - epochStart;
         grossSettledAmount = rate * duration;
-        settledUntilEpoch = epochEnd;
+        uint256 settledUntilEpoch = epochEnd;
         note = "";
 
         // If this rail has an validator and we're not skipping validation, let it decide on the final settlement amount


src/FilecoinPayV1.sol
```diff
@@ -1529,6 +1518,30 @@ contract FilecoinPayV1 is ReentrancyGuard {
         );
     }
 
+    function _settleSegment(uint256 railId, uint256 epochStart, uint256 epochEnd, uint256 rate, bool skipValidation)
```
wjmelements: is this still being used?

src/FilecoinPayV1.sol
```diff
@@ -1352,61 +1348,65 @@ contract FilecoinPayV1 is ReentrancyGuard {
...
-                    state.totalNetworkFee,
-                    validationNote
-                );
+                // Apply fees to whatever we've settled so far
+                (totalNetPayeeAmount, totalOperatorCommission, totalNetworkFee) =
+                    calculateAndPayFees(totalGrossSettled, rail.token, rail.serviceFeeRecipient, rail.commissionRateBps);
+                payee.funds += totalNetPayeeAmount;
+                totalSettledAmount = totalGrossSettled;
+                return
+                    (totalSettledAmount, totalNetPayeeAmount, totalOperatorCommission, totalNetworkFee, validationNote);
```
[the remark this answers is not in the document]
Chaitu-Tatipamula: ah yes, we can break at early exits

src/FilecoinPayV1.sol
Chaitu-Tatipamula: yes exactly so then we get rid of `_settleSegment ` altogether?

src/FilecoinPayV1.sol
```diff
@@ -1323,25 +1323,21 @@ contract FilecoinPayV1 is ReentrancyGuard {
     {
         Rail storage rail = rails[railId];
         RateChangeQueue.Queue storage rateQueue = rail.rateChangeQueue;
+        Account storage payee = accounts[rail.token][rail.to];
```
wjmelements: I think the code gets a little simpler if you calculate the `payee` in `settleRailInternal` too

src/FilecoinPayV1.sol
wjmelements: Yes it can be replaced by `_settleSegmentGross`.

src/FilecoinPayV1.sol
```diff
@@ -1352,61 +1356,46 @@ contract FilecoinPayV1 is ReentrancyGuard {
...
+        // Apply fees once after loop completion (whether normal completion or early exit)
+        if (totalGrossSettled > 0) {
+            (totalNetPayeeAmount, totalOperatorCommission, totalNetworkFee) =
+                calculateAndPayFees(totalGrossSettled, rail.token, rail.serviceFeeRecipient, rail.commissionRateBps);
+            payee.funds += totalNetPayeeAmount;
+        } else {
+            totalNetPayeeAmount = 0;
+            totalOperatorCommission = 0;
+            totalNetworkFee = 0;
+        }
```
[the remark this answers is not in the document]
Chaitu-Tatipamula: yes yes about to do it

wjmelements (approved):

PR #271, opened 2026-01-28
fix(settlement): accumulate validator notes across rate change segments

### LINKED ISSUE #268: Audit Fix L02: Accumlate notes
Audit reference: `[FIL-1132b525-L02`

When settling a queue with multiple rate changes only the last note is kept.  We should do something like in [finalization](https://github.com/FilOzone/filecoin-pay/blob/1132b525e623e5f459e520121627ed2a159d33c9/src/FilecoinPayV1.sol#L1291) passing the note through and concatenating.

I am a bit worried about memory / ret value / processing costs adding up quickly.  Maybe we can concat only if strings don't match? Or perhaps enforce really strict limits on note size?

ZenGround0:
Closes #268 

I went with the append option.  I think this is fine as non empty notes only show up as error conditions in our only existing validator and as auditors have pointed out settling many segments will cost linearly increasing gas so this is only adding a constant term to an existing efficiency problem not creating a whole new problem.  And it is a simple matter for users to settle smaller spans of time if we actually start hitting gas problems.

### DIFF
--- a/src/FilecoinPayV1.sol
+++ b/src/FilecoinPayV1.sol
@@ -1361,7 +1361,15 @@ contract FilecoinPayV1 is ReentrancyGuard {
 
             // Add the gross settled amount to our running total
             totalGrossSettled += segmentGrossSettled;
-            note = validationNote;
+
+            // Accumulate validation notes from each segment
+            if (bytes(validationNote).length > 0) {
+                if (bytes(note).length > 0) {
+                    note = string.concat(note, "; ", validationNote);
+                } else {
+                    note = validationNote;
+                }
+            }
 
             // If validator partially settled the segment, exit early
             if (rail.settledUpTo < segmentEndBoundary) {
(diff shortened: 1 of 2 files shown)


src/FilecoinPayV1.sol
```diff
@@ -1361,7 +1361,15 @@ contract FilecoinPayV1 is ReentrancyGuard {
 
             // Add the gross settled amount to our running total
             totalGrossSettled += segmentGrossSettled;
-            note = validationNote;
+
+            // Accumulate validation notes from each segment
+            if (bytes(validationNote).length > 0) {
+                if (bytes(note).length > 0) {
+                    note = string.concat(note, "; ", validationNote);
```
wjmelements: this is a pretty expensive operation in ethereum because memory gas is quadratic and solidity will leak each of these strings, but it's probably fine in the FEVM.

PR #272, opened 2026-01-28
fix(RateChangeQueue): empty queue behavior

### LINKED ISSUE #266: Audit Fix 1: RateChangeQueue clearing behavior
PR #86 introduced unreachable code.  Upon dequeue of the rate change queue down to zero elements we try to reset the array completely with some fancy evm assembly zeroing out the changes dynamic array slot.

The reason for this as far as I can tell is code cleanliness.  Super-theoretically it makes overflowing the head and changes.length slots easier to reason about.  But reasonably-theoretically we could never overflow these 256 bit ints if we ran for the age of the universe.  

Either removing this case entirely and letting the array's length grow without resetting to zero OR doing head++ first and then checking isEmpty condition are acceptable.  I lean towards just letting array grow to remove a tiny bit of code from the expensive settlement operation.

Note that if KAMT behaved like the naive AMT datastructure used in other places in filecoin actors there would be a significant state overhead to operating with larger array sizes.  There is a known inefficiency where the AMT adds a greater number of internal nodes eagerly when adding a bigger integer key than a small integer key.  This is not the case with the HAMT which symmetrically handles all keys since all keys are hashed.  The KAMT is a modification of the core HAMT logic and does not suffer from this inefficiency.

wjmelements:
Fixes #266
Reviewer @zenground0
#### Changes
* move reset of head = 0 to when zeroing rail
* change string revert error 'Queue is empty' into abi 'error EmptyQueue()' 
* remove redundant bounds checking with `unchecked`

### DIFF
--- a/src/FilecoinPayV1.sol
+++ b/src/FilecoinPayV1.sol
@@ -1596,6 +1596,7 @@ contract FilecoinPayV1 is ReentrancyGuard {
         rail.settledUpTo = 0;
         rail.endEpoch = 0;
         rail.commissionRateBps = 0;
+        rail.rateChangeQueue.clearEmpty();
     }
 
     function updateOperatorRateUsage(OperatorApproval storage approval, uint256 oldRate, uint256 newRate) internal {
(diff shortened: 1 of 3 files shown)


ZenGround0:
Looks like we could take the last two bullet points as nice add ons in addition to PR #269

